Issue #25, opened 2026-05-16
[BUG] macOS client sends message when pressing Enter to confirm IME candidate


ve9etandy:
## Description

In the Suzent macOS client, when typing in the input box with a Chinese IME, pressing `Enter` while selecting an IME candidate sends the message immediately instead of committing the selected candidate text into the input box.

For example, when I type `ni hao`, the Chinese IME candidate list appears. I expect pressing `Enter` to select/commit `你好`, but the message is sent directly instead.

<img width="1670" height="966" alt="Image" src="https://github.com/user-attachments/assets/3ee3660c-2f4c-4ab0-9119-ce73ea14b7ee" />

## Expected Behavior

When an IME composition is active, pressing `Enter` should first be handled by the input method to confirm the selected candidate text.

The message should only be sent after the IME composition has ended and the user presses `Enter` again.

## Actual Behavior

Pressing `Enter` during IME candidate selection triggers the send action directly, before the selected Chinese text is committed to the input box.

## Environment

- Platform: macOS
- App: Suzent macOS client
- Input method: Chinese Pinyin IME

## Possible Cause

The input box may be handling the `Enter` key without checking whether an IME composition is currently active.

## Steps to Reproduce

1. Open the Suzent macOS client.
2. Click the message input box.
3. Switch to a Chinese input method, such as macOS Pinyin.
4. Type `ni hao`.
5. While the IME candidate list is still active, press `Enter`.
6. The message is sent immediately.

cyzus:
#28

PR #26, opened 2026-05-16
fix: enhance input handling in ChatInputPanel to support IME composit…

ve9etandy:
https://github.com/cyzus/suzent/issues/25


cyzus:
fixed in #28

PR #29, opened 2026-05-17
Refactor static config out of chat database

### LINKED ISSUE #24: Refactor: Extract User Preferences and Secrets from SQLite Database for Better Syncing
## Problem
Currently, Suzent stores configuration (`user_preferences`) and secrets (`api_keys`) within the SQLite database (`chats.db`). The `api_keys` table stores secrets using either Keyring or an Encrypted DB backend with Fernet (via `src/suzent/core/secrets.py`).

While encryption is handled correctly, storing these configurations alongside highly volatile runtime data (like `chats` and `tasks`) creates significant friction for users who want to synchronize their agent's capabilities across multiple devices using tools like Syncthing or GitHub.

Users face a dilemma:
1. **Sync `chats.db`**: Leads to database corruption and sync conflicts when multiple instances run simultaneously.
2. **Don't sync `chats.db`**: Forces users to manually reconfigure API keys, preferred models, and tool whitelists on every device.

## Proposed Solution
Decouple static capabilities (config/secrets) from runtime state (history/tasks) following 12-Factor App principles:

1. **Extract Secrets**: Shift the source of truth for API keys entirely to environment variables (`.env` file) or the OS Keyring, rather than maintaining an `api_keys` table inside the shared database.
2. **Extract User Preferences**: Move `user_preferences` and `memory_config` to a plain text configuration file (e.g., `~/.suzent/config.yaml`).
3. **Isolate Database**: Ensure `chats.db` contains only runtime state (chat history, task graphs) so it can be safely added to `.gitignore` or Syncthing ignore lists without losing the agent's core capabilities across devices.

## Affected Areas
- `src/suzent/database.py` (`ApiKeyModel`, `UserPreferencesModel`)
- `src/suzent/core/secrets.py`
- Any components relying on `db.get_api_keys()` or `db.get_user_preferences()`

HxyScotthuang:
## Summary
- Move user preferences, memory config, and provider config blobs out of chats.db into user-scoped YAML config
- Store secrets through SecretManager using OS keyring by default, with a Fernet-encrypted SQLite fallback in a separate secrets.db
- Add conservative legacy migration that moves api_keys, user_preferences, and memory_config rows out of chats.db and only drops legacy tables after successful transfer
- Keep existing ChatDatabase APIs as compatibility shims for current call sites

## Scope note
This addresses #24 as a prerequisite for syncing agent capabilities across devices. It does not implement a central shared brain/memory database or distributed concurrency model; runtime state remains local.

Closes #24

## Testing
- uv run --extra dev ruff check src\suzent\core\secrets.py src\suzent\core\user_config.py src\suzent\database.py tests\core\test_secrets.py tests\integration\test_database.py
- uv run --extra dev pytest tests\core\test_secrets.py tests\integration\test_database.py -q
- uv run --extra dev pytest -m "not sandbox" -q


src/suzent/core/user_config.py
```diff
@@ -0,0 +1,103 @@
...
+from datetime import datetime
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+from suzent.config import DATA_DIR
+
+
+def get_user_config_path() -> Path:
```
cyzus, line 16: USER_CONFIG_DIR

PR #35, opened 2026-05-19
fix(providers): default chat model picks first configured provider (#33)

### LINKED ISSUE #33: Default chat model should use first configured provider, not Anthropic catalog default
## Summary

On a fresh install (no `_PROVIDER_CONFIG_` in the DB), Suzent exposes **all** catalog default models and picks `models[0]` as the default chat model. Because the list is **alphabetically sorted**, that is always an Anthropic model (`anthropic/claude-haiku-4-5`), even when the user has only configured another provider (e.g. OpenAI, Gemini, Ollama).

The agent **persona** is not the problem — `agent_options` is a single entry (Agent`). The UX issue is the default **LLM model / provider**.

## Current behavior (verified)

1. **`get_enabled_models_from_db()`** (`src/suzent/core/providers/helpers.py`): when there is no provider config blob, it returns every `default_models` id from `config/providers.json`, then `sorted(set(...))`.
2. **`/config` API** exposes that list as `models`.
3. **Frontend** (`frontend/src/hooks/useChatStore.tsx`) sets `model: backendConfig.models[0]` when there are no saved user preferences.
4. Result: first model is **`anthropic/claude-haiku-4-5`** (alphabetically first among 24 catalog defaults).

```python
# helpers.py — fallback when no _PROVIDER_CONFIG_
defaults = [m[id] for p in PROVIDER_REGISTRY for m in p.default_models]
return sorted(set(defaults))  # → anthropic/* first
```

```typescript
// useChatStore.tsx
model: prefs?.model || backendDefaults.models[0] || '',
```

## Expected behavior

1. **Default model** should be the first model from the **first provider that is actually available** (has credentials / verified connection), in a stable provider order (e.g. catalog order in `providers.json`: OpenAI → ChatGPT subscription → Anthropic → …), not global alphabetical sort across all providers.
2. If **no** provider is configured, do not silently default to Anthropic — either:
   - show an empty / disabled model selector with a clear “configure a provider” prompt, or
   - only list models for providers with keys set (env or Settings).
3. **New chat / first-screen UX**: on the new-chat view (model selector in `NewChatView` / `ChatInputPanel`), show which providers are configured vs not (e.g. badges or grouped dropdown: “OpenAI ✓”, “Anthropic — not configured”). Reuse or extend `GET /config/api-keys` provider `isSet` status already returned for Settings.

### Backend
- Add something like `get_default_chat_model()` that:
  1. Iterates `PROVIDER_REGISTRY` in catalog order.
  2. For each provider, checks `resolve_api_key(provider.id)` (and ChatGPT subscription auth where applicable).
  3. Returns the first `enabled_models` entry for that provider, or first discovered/default model after verify.
- Change `get_enabled_models_from_db()` to either:
  - return only models for configured providers, **or**
  - return grouped metadata `{ provider_id, configured, models[] }` so the UI can render status.
- Include `default_model` (and optionally `default_model_provider`) in `GET /config` response.

### Frontend
- Use `default_model` from backend instead of `models[0]`.
- On **new chat** screen: show provider configuration status (configured / missing key / not verified) before or beside the model dropdown.
- Optional: prompt to open Settings → Providers when no provider is configured.

## Files involved

- `src/suzent/core/providers/helpers.py` — model list + default selection
- `src/suzent/routes/config_routes.py` — `/config` payload
- `frontend/src/hooks/useChatStore.tsx` — default config bootstrap
- `frontend/src/components/NewChatView.tsx` / `ChatInputPanel.tsx` — first-screen model UI
- `frontend/src/components/settings/ProvidersTab.tsx` — reference for provider status patterns

## Acceptance criteria

- [ ] Fresh install with only `OPENAI_API_KEY` set defaults to an OpenAI model, not Anthropic.
- [ ] Fresh install with **no** keys does not pre-select Anthropic; user is guided to configure a provider.
- [ ] New-chat UI indicates which providers are configured vs not.
- [ ] Existing saved user preferences continue to take precedence.

## Notes

- Background model refresh (`server.py` `_refresh_provider_models`) already walks providers in registry order and skips those without keys — similar logic should drive the default model choice.
- Clarification: “agent” in the UI refers to the single Suzent agent type; this issue is about **model/provider** defaults.

chiruno-9:
## Summary

Fixes #33 — on a fresh install with only one provider configured (e.g. only `OPENAI_API_KEY` set), the new-chat view always landed on `anthropic/claude-haiku-4-5` because `get_enabled_models_from_db()` returns every catalog default sorted alphabetically and the frontend used `models[0]`.

## Changes

**Backend**
- `src/suzent/core/providers/helpers.py` — add `get_default_chat_model()`:
  - Walks `PROVIDER_REGISTRY` in catalog order (`providers.json`: OpenAI → ChatGPT → Anthropic → Gemini → …).
  - For each provider: API-keyed providers checked via `resolve_api_key()`; the keyless ChatGPT subscription provider checked via `is_authenticated()`.
  - For the first configured provider, returns the user's first `enabled_models` entry (from `_PROVIDER_CONFIG_`) or falls back to `default_models[0]`.
  - Returns `None` when nothing is configured so callers can distinguish "unconfigured" from "first item in some list".
- `src/suzent/routes/config_routes.py` — exposes the result as `defaultModel` on `/config`.

**Frontend**
- `frontend/src/types/api.ts` — `ConfigOptions.defaultModel?: string | null`.
- `frontend/src/hooks/useChatStore.tsx` — both the preferences-merge path and the no-localStorage fallback now use `defaultModel ?? models[0]`. The optional field keeps the change backward-compatible with older backends.

**Tests**
- `tests/core/test_default_chat_model.py` (7 tests) covers:
  - no provider configured → `None`,
  - only OpenAI configured → OpenAI model (not Anthropic),
  - catalog order is respected when an earlier provider is unconfigured,
  - user `enabled_models` take precedence over catalog defaults,
  - empty user `enabled_models` falls back to catalog defaults,
  - `_provider_is_configured` uses `resolve_api_key`,
  - malformed `_PROVIDER_CONFIG_` blob → `{}` (no crash).

## Out of scope (deliberately deferred)

The issue also proposes a UI status indicator (provider configured / not-configured badges on the new-chat screen). That belongs in a separate PR — this one is scoped to the data-loss-style default-selection bug so it can be reviewed and merged quickly.

## Acceptance criteria (issue #33)

- [x] Fresh install with only `OPENAI_API_KEY` set defaults to an OpenAI model, not Anthropic.
- [x] Fresh install with no keys → `defaultModel` is `null`; the existing `models[0]` UX is unchanged (UI prompt for provider config can land in a follow-up).
- [ ] New-chat UI indicates which providers are configured vs not. *(deferred to follow-up)*
- [x] Existing saved user preferences continue to take precedence (`prefs?.model` is still checked first in `buildConfigFromPreferences`).

## Test plan

- [x] `uv run pytest tests/core/test_default_chat_model.py` — 7/7 pass.
- [x] `uv run pytest tests/core/test_provider_registry.py` — 18/18 pass (no regression).
- [x] `uv run ruff check` on changed files — clean.
- [x] `npx tsc --noEmit` in `frontend/` — clean.
- [x] `npx vitest run` — 21/21 pass.
- [ ] Manual smoke: fresh install with only `OPENAI_API_KEY` set → new-chat selector lands on `openai/gpt-4.1`.

### DIFF
--- a/frontend/src/hooks/useChatStore.tsx
+++ b/frontend/src/hooks/useChatStore.tsx
@@ -91,7 +91,7 @@ const buildConfigFromPreferences = (
   prefs: ConfigOptions['userPreferences'],
   backendDefaults: ConfigOptions
 ): ChatConfig => ({
-  model: prefs?.model || backendDefaults.models[0] || '',
+  model: prefs?.model || backendDefaults.defaultModel || backendDefaults.models[0] || '',
   agent: prefs?.agent || backendDefaults.agents[0] || '',
   tools: prefs?.tools || backendDefaults.defaultTools || [],
   memory_enabled: prefs?.memory_enabled,
@@ -298,7 +298,7 @@ export const ChatProvider: React.FC<{ children: React.ReactNode; enabled?: boole
     // Fallback to backend defaults
     if (backendConfig) {
       return {
-        model: backendConfig.models[0] || '',
+        model: backendConfig.defaultModel || backendConfig.models[0] || '',
         agent: backendConfig.agents[0] || '',
         tools: backendConfig.defaultTools || [],
         sandbox_enabled: backendConfig.sandboxEnabled ?? true,
--- a/frontend/src/types/api.ts
+++ b/frontend/src/types/api.ts
@@ -107,6 +107,7 @@ export interface Plan {
 export interface ConfigOptions {
   title: string;
   models: string[];
+  defaultModel?: string | null; // first model from a provider with credentials; null if none configured
   agents: string[];
   tools: string[];        // full list of tool options
   defaultTools: string[]; // default enabled tools
--- a/src/suzent/core/providers/__init__.py
+++ b/src/suzent/core/providers/__init__.py
@@ -14,6 +14,7 @@ from suzent.core.providers.catalog import (
 from suzent.core.providers.factory import ProviderFactory
 from suzent.core.providers.generic import GenericLiteLLMProvider
 from suzent.core.providers.helpers import (
+    get_default_chat_model,
     get_effective_memory_config,
     get_enabled_models_from_db,
     resolve_api_key,
@@ -49,5 +50,6 @@ __all__ = [
     "ProviderFactory",
     "resolve_api_key",
     "get_enabled_models_from_db",
+    "get_default_chat_model",
     "get_effective_memory_config",
 ]
--- a/src/suzent/core/providers/helpers.py
+++ b/src/suzent/core/providers/helpers.py
@@ -137,24 +137,38 @@ def get_default_chat_model() -> Optional[str]:
     """Return the default chat model id from the first configured provider.
 
     Walks ``PROVIDER_REGISTRY`` in catalog order and returns the first model
-    available from a provider that has credentials. The user's
-    ``enabled_models`` list (from ``_PROVIDER_CONFIG_``) takes precedence over
-    the catalog ``default_models`` for the chosen provider.
+    available from a provider that has credentials.
 
-    Returns ``None`` if no provider is configured, so callers (UI / API) can
-    distinguish "no setup" from "first model in some list".
+    Semantics must match ``get_enabled_models_from_db`` so the returned id is
+    always a member of ``/config.models``:
+
+    * **No ``_PROVIDER_CONFIG_`` blob saved** (fresh install): fall back to the
+      provider's catalog ``default_models[0]``.
+    * **Blob saved**: the blob is the source of truth. Honor each provider's
+      ``enabled_models`` strictly — an explicit empty list, or no entry at
+      all, means "this provider exposes no chat models" and we walk on. We do
+      *not* fall back to catalog defaults in this case, otherwise the helper
+      could return a model that the frontend cannot select.
+
+    Returns ``None`` if no provider is configured.
     """
     from suzent.core.providers.catalog import PROVIDER_REGISTRY
 
     user_provider_config = _load_user_provider_config()
+    blob_present = bool(user_provider_config)
 
     for spec in PROVIDER_REGISTRY:
         if not _provider_is_configured(spec):
             continue
 
-        enabled = user_provider_config.get(spec.id, {}).get("enabled_models") or []
-        if enabled:
-            return enabled[0]
+        if blob_present:
+            # Strictly honor the saved blob — matches get_enabled_models_from_db.
+            enabled = user_provider_config.get(spec.id, {}).get("enabled_models") or []
+            if enabled:
+                return enabled[0]
+            continue
+
+        # Fresh install: blob absent, fall back to catalog defaults.
         if spec.default_models:
             return spec.default_models[0]["id"]
 
--- a/src/suzent/core/providers/helpers.py
+++ b/src/suzent/core/providers/helpers.py
@@ -89,6 +89,78 @@ def get_enabled_models_from_db() -> List[str]:
     return sorted(set(all_models))
 
 
+def _load_user_provider_config() -> Dict[str, Any]:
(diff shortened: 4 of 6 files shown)


HxyScotthuang:
Thanks for the fix. I tested this locally and the main happy path looks good:

- `uv run pytest tests/core/test_default_chat_model.py tests/core/test_provider_registry.py` passed
- `uv run ruff check` on the changed backend files passed
- `npm test -- --run` passed
- `npm run build` passed, with only existing Vite chunk warnings

One issue I think should be fixed before approval:

`get_default_chat_model()` can return a model that is not present in `/config.models`.

If `_PROVIDER_CONFIG_` exists and an earlier configured provider has an explicit empty `enabled_models` list, the new helper falls back to that provider’s catalog default:

```py
enabled = user_provider_config.get(spec.id, {}).get(enabled_models) or []
if enabled:
    return enabled[0]
if spec.default_models:
    return spec.default_models[0][id]
```

But `get_enabled_models_from_db()` only returns models from saved `enabled_models` when `_PROVIDER_CONFIG_` exists. So in a case like:

```py
{
    openai: {enabled_models: []},
    anthropic: {enabled_models: [anthropic/claude-haiku-4-5]},
}
```

with both providers configured, `defaultModel` becomes `openai/gpt-4.1`, while `models` only contains `anthropic/claude-haiku-4-5`.

That means the frontend can initialize to a model that is not selectable, and backend agent creation may reject it and silently fall back to `enabled_models[0]`.

Could we distinguish “provider has no saved config yet” from “provider exists and intentionally has no enabled models”? I think catalog fallback should only happen for the no-saved-entry case, not for an explicit empty `enabled_models`.

chiruno-9:
Great catch — you're right, and the bug is exactly as you described.

The fix in 0e4fbbf: when ``_PROVIDER_CONFIG_`` is present, the blob is the source of truth — providers with no entry or an explicit empty ``enabled_models`` are skipped instead of falling back to catalog defaults. Catalog fallback now only fires when the blob is entirely absent (fresh install), which matches the branch in ``get_enabled_models_from_db()``.

Added four tests pinning the cases:

- ``test_explicit_empty_enabled_models_skips_provider_when_blob_present`` — your exact scenario: openai has ``enabled_models: []``, anthropic has a list → returns the anthropic model.
- ``test_provider_configured_but_missing_from_blob_is_skipped`` — provider has credentials but no blob entry → still skipped.
- ``test_returns_none_when_blob_present_but_all_enabled_lists_empty`` — degenerate case.
- ``test_falls_back_to_catalog_default_only_when_blob_absent`` — regression for the original fresh-install fix.

All 10 tests pass; ruff clean. Ready for another look when you have a moment.

cyzus:
It looks like the newly added `_load_user_provider_config()` function duplicates the database fetching and JSON parsing logic already present in `get_enabled_models_from_db()`. 

We can refactor `get_enabled_models_from_db()` to reuse this new helper to keep the code DRY.

Here is a suggestion for `src/suzent/core/providers/helpers.py`:

```python
def get_enabled_models_from_db() -> List[str]:
    """Aggregate all enabled/available models from user provider config."""
    from suzent.config import CONFIG
    
    custom_config = _load_user_provider_config()

    if not custom_config:
        if CONFIG.model_options:
            return CONFIG.model_options
        from suzent.core.providers.catalog import PROVIDER_REGISTRY

        defaults = [m["id"] for p in PROVIDER_REGISTRY for m in p.default_models]
        return sorted(set(defaults))

    from suzent.core.providers.catalog import PROVIDER_REGISTRY

    all_models = [
        model_id
        for p in PROVIDER_REGISTRY
        for model_id in custom_config.get(p.id, {}).get("enabled_models", [])
    ]

    return sorted(set(all_models))
```

cyzus:
#39

PR #54, opened 2026-06-18
feat(tools): session_search tool with FTS5-backed chat search

cyzus:
## Summary

Adds a **`session_search`** tool so the agent can recall and navigate past chat
sessions, plus the shared **FTS5 full-text search infrastructure** it runs on — which
the existing chat-list search now uses too.

Ports the behavior of Hermes-agent's `session_search` (one tool, mode dispatch by which
params are set), adapted to this codebase's storage model.

## Tool modes

| Mode | Trigger | Returns |
|---|---|---|
| **Read** | `session_id` | One whole session, sanitized (first 20 + last 10 if large) |
| **Discovery** | `query` | Top-N matching sessions with snippet, ±context window, bookends |
| **Browse** | neither | Recent sessions, chronological |

`role_filter` (default `user,assistant`) gates tool output; the current session is
excluded from Discovery/Browse (recall is about *other* conversations).

## Infrastructure (`ChatSearchMixin`)

- New `chat_messages_fts` FTS5 virtual table, **trigram tokenizer** so both ASCII and
  CJK substring search work (`unicode61` can't match a substring of an undelimited CJK
  run). Queries below the 3-char trigram minimum fall back to the prior `LIKE` scan, so
  short/CJK **chat-list search never regresses**.
- Messages are stored as a JSON blob, not row-per-message, so FTS5 sync triggers can't
  be used — instead the index is **reindexed at the application layer on every chat
  write path** (`create`/`update`/`append`/`finalize`), with cleanup on delete.
- Idempotent table creation + one-time backfill in migrations. If the SQLite build lacks
  FTS5, everything **degrades to `LIKE`** and chat writes are never blocked.

## Sanitization

All output comes from each message's clean **AG-UI `parts`**, never the legacy HTML
`content` string. Assistant **reasoning is never surfaced**; tool output only appears
when `role_filter` includes `tool`.

## Tests

`tests/integration/test_chat_search.py` — sanitization (reasoning dropped, tool gated,
HTML ignored), reindex on each write path + delete, Discovery/Read DB methods, hidden-
platform exclusion, chat-list FTS parity + CJK `LIKE` fallback, and the three tool modes.
All green; existing DB / chat-processor / chat-route suites still pass.

## UI grouping & equipment

- New **"Memory & recall"** tool group (`ToolGroup.MEMORY`) containing `session_search`
  and `memory_search`. The frontend renders it automatically from the backend
  tool-groups payload — no UI change required.
- **`memory_search` is now a sidebar-toggleable equipment tool**, decoupled from the
  global memory toggle. It is equipped via the `tools` list like any other tool;
  `memory_enabled` now governs only memory *context injection*, not tool availability.
- Both tools added to `default_tools`, so recall works out of the box and existing
  memory users keep `memory_search` after the decoupling.

## Tests (cont.)

`tests/core/test_memory_tool_equipment.py` — memory_search equipped when in `tools`
regardless of `memory_enabled`, and absent when not in `tools` even with memory enabled.

## Notes

- `config/capabilities/dashscope.json` was modified before this work and is intentionally
  left out of this branch.

### DIFF
--- a/frontend/src/components/chat/ToolsPanel.tsx
+++ b/frontend/src/components/chat/ToolsPanel.tsx
@@ -44,9 +44,8 @@ export function ToolsPanel(): React.ReactElement | null {
     <div className="flex items-center justify-center h-full text-neutral-400 text-xs">Loading…</div>
   );
 
-  const EXCLUDED = ['MemorySearchTool', 'MemoryBlockUpdateTool'];
   const rawTools = Array.isArray(backendConfig.tools) ? backendConfig.tools : [];
-  const available: string[] = rawTools.filter((t: string) => !EXCLUDED.includes(t));
+  const available: string[] = rawTools;
   const selected: string[] = config.tools || [];
 
   const sourceGroups: { label: string; tools: string[] }[] = backendConfig.toolGroups ?? [];
--- a/frontend/src/components/settings/AutomationTab.tsx
+++ b/frontend/src/components/settings/AutomationTab.tsx
@@ -258,7 +258,6 @@ export function AutomationTab({ models, tools = [] }: AutomationTabProps): React
                   saveHeartbeatGlobalConfig({ allowed_tools: newTools });
                 }}
                 options={tools
-                  .filter(t => !['MemorySearchTool', 'MemoryBlockUpdateTool'].includes(t))
                   .map(t => ({
                     value: t,
                     label: t.replace(/Tool$/, '').replace(/([a-z])([A-Z])/g, '$1 $2').toUpperCase(),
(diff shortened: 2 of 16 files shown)


cyzus:
Thanks for the review — addressed all four in `49dca5c`:

- **P1 (legacy content leak):** A parts-less assistant row now returns empty instead of its raw `content`. Assistant text only ever comes from `parts`; the legacy `content` string (reasoning `<details>`, a2ui `<div>`) is never surfaced. Test: `test_legacy_assistant_without_parts_never_leaks_content`.
- **P2 (Discovery LIKE fallback):** `search_chat_messages` now falls back to a LIKE scan (`messages_search_filter`) when FTS can't serve the query — no FTS5, sub-trigram/CJK queries, or invalid FTS syntax — matching the chat-list degradation. Test: `test_discovery_falls_back_to_like_for_short_query`.
- **P2 (current-chat exclusion):** Excluded via `exclude_chat_id` inside `search_chat_messages`, before the limit is applied, so it never consumes a result slot. Test: `test_discovery_excludes_chat_before_limit`.
- **P2 (resumable backfill):** Backfill now indexes only chats with no rows in the index (instead of skipping everything when the index is non-empty), so an interrupted first backfill completes on the next startup. Test: `test_backfill_resumes_for_unindexed_chats`.

PR #55, opened 2026-06-18
fix(db): keep sidebar message summary consistent across rollbacks

cyzus:
## Summary

Fixes the stale **sidebar message count** bug surfaced during the session_search work
(#54), where a chat could show e.g. "0 messages" despite having replies.

## Root cause

The cached sidebar summary (visible count + preview) lives in `chat.config`.
`list_chats` trusts the stored value and only recomputes it when it's **missing**
([chats.py](../blob/main/src/suzent/database/chats.py)). Every write path *in* `chats.py`
refreshes the summary — but two rollback paths rewrote `chat.messages` **directly**,
bypassing the refresh:

- retry checkpoint restore — `core/retry.py`
- heartbeat no-op rollback — `core/heartbeat.py`

After either ran, the stored summary kept its pre-rollback value forever.

## Fix

- New `ChatOperationsMixin.rewrite_chat_messages()` — replaces the message list while
  refreshing the summary **and** the FTS index (and optionally `agent_state` /
  `turn_count`). Both rollback paths now route through it instead of assigning
  `chat.messages` raw.
- One-time startup repair `_repair_stale_chat_summaries()` — recomputes every chat's
  summary once and persists any that drifted, guarded by a config-blob flag so it runs
  a single time. (Verified it corrected the real stale session from #54: 0 → 3.)

The count **metric** is unchanged (assistant replies with visible text); only its
staleness is fixed.

## Tests

`tests/integration/test_chat_summary_consistency.py` — summary refresh on rewrite,
`turn_count` delta, list_chats consistency after rollback, the one-time repair fixing a
stale count, and the repair being a no-op on second run. Existing DB / search /
heartbeat / postprocess suites still pass.

### DIFF
--- a/src/suzent/core/heartbeat.py
+++ b/src/suzent/core/heartbeat.py
@@ -367,21 +367,19 @@ class HeartbeatRunner(BaseBrain):
     def _rollback_heartbeat_messages(self, chat_id: str, original_count: int, db):
         """Remove the heartbeat prompt and Ok response if no action was needed."""
         try:
-            from sqlmodel import Session
-            from sqlalchemy.orm.attributes import flag_modified
-            from suzent.database import ChatModel
-
-            with Session(db.engine) as session:
-                chat = session.get(ChatModel, chat_id)
-                if chat and len(chat.messages) > original_count:
-                    extra = len(chat.messages) - original_count
-                    chat.messages = chat.messages[:original_count]
-                    chat.turn_count = max(0, (chat.turn_count or 0) - 1)
-                    flag_modified(chat, "messages")
-                    session.commit()
-                    logger.debug(
-                        f"Rolled back {extra} heartbeat messages from chat {chat_id}"
-                    )
+            chat = db.get_chat(chat_id)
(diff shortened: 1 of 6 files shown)


cyzus:
Thanks — addressed both in `991e863`:

- **P1 (restore NULL agent_state):** `rewrite_chat_messages` now uses an `_UNSET` sentinel for `agent_state`, so omitting it (heartbeat rollback) leaves stored state untouched, while passing `agent_state=None` (a first-turn retry checkpoint) actually restores NULL. Retry passes it explicitly. Tests: `test_rewrite_can_restore_null_agent_state`, `test_rewrite_omitting_agent_state_leaves_it_untouched`.
- **P2 (DB-local marker):** the one-time repair marker now lives in a `schema_meta` table inside chats.db (via `_get/_set_schema_meta`) instead of the synced config.yaml, so a marker from one device can't suppress repair of another device's independent chats.db.

cyzus (commented):
lgtm

PR #139, opened 2026-08-26
fix: address repository context review feedback

cyzus:
## Summary
- scope loaded repository context and request completion to the active chat
- return repository instructions even when memory is disabled or unavailable
- add the missing Tools tab translations in both locales
- show localized project-context save errors while keeping the editor open

Addresses the delayed automated review feedback on #136 and #137.

## Validation
- frontend build
- 206 frontend tests
- 1,475 non-sandbox backend tests
- 6 focused route tests
- formatting and lint checks


cyzus:
@codex review

PR #162, opened 2026-09-02
fix(security): make system-reminder boundaries unforgeable

cyzus:
## Problem

The delimiters marking hidden reminder blocks — `U+E203`/`U+E204` and `<system-reminder>` — are public constants. Any text that reaches the model could carry a byte-identical block and be read as trusted out-of-band context.

Tool output is the sharp end. A fetched web page, a file read, or an MCP server response is attacker-influenceable, and `STATIC_INSTRUCTIONS` explicitly tells the model to trust these blocks and act on them without mentioning them to the user. That is a prompt-injection path with the user kept in the dark.

A second, smaller issue: `strip_system_reminders()` hid *any* look-alike block from the UI, including one a user typed themselves — so forged content could be visible to the model and invisible in the transcript.

## What this does not change

Reminders stay in `UserPromptPart`. That placement is deliberate and correct: the system prompt is the prompt-cache prefix and must stay stable, so per-turn volatile context (RAG hits, plan state, skill catalog) belongs at the tail of the user message. Moving reminders to a system-role message would invalidate the cached prefix every single turn. This PR fixes the forgeable boundary without touching the layout.

## Approach

1. **Per-process token.** Every reminder we author embeds `RUNTIME_NONCE`, so a genuine block cannot be produced by anything that cannot read this process's memory.
2. **Ingress sanitizing** (`sanitize_untrusted_text`) on user messages and attachment text, before the genuine reminder is appended.
3. **Tool results** go through a history processor — the only chokepoint, since pydantic-ai builds `ToolReturnPart` internally. Registered *before* the compaction processor so forged delimiters can't slip through folded into a summary.

Delimiters are replaced with a visible `[reminder-delimiter]` marker rather than deleted: the text stays intelligible, the attempt stays auditable, and nothing the user wrote disappears.

The token is defence in depth, not the primary control — it lands in persisted history, so anyone who can read a raw transcript can learn it. Sanitizing is what actually holds. The code says so where it matters.

## One coupling worth flagging in review

Display still strips *untokenized* blocks, because histories written before this change have none, and those reminders must keep staying hidden. That is only safe because ingress sanitizing guarantees a surviving delimiter is runtime-authored. If someone later removes the ingress step, `strip_system_reminders()` silently starts hiding text users actually typed. The docstring records this; a reviewer may reasonably want it enforced by something stronger than a comment.

## Tests

9 new cases in `tests/core/test_system_reminder.py`: forged PUA and XML blocks at ingress, forged blocks in tool output, correct-token forgery, user-typed delimiters staying visible, legacy and other-process reminders staying hidden, and clean tool output plus genuine reminders passing through untouched.

Full suite: 1647 passed. `ruff check` and `ruff format` clean.

## Context

From the P0 in `docs/03-developing/system-prompt-and-reminder-audit.md`. A follow-up PR covers P1-0 (cached agents serving stale core memory).

### DIFF
--- a/src/suzent/acp/runtime.py
+++ b/src/suzent/acp/runtime.py
@@ -236,6 +236,16 @@ async def stream_acp_turn(
     file_context = _build_acp_file_context(file_mentions, files)
     if file_context:
         message = f"{file_context}\n\n{message}" if message else file_context
+        # File annotations interpolate caller-supplied paths, so the assembled
+        # prompt is untrusted again even though `message` was already clean.
+        # Sanitizing the finished string rather than only its parts is the point:
+        # anything later prepended or appended here is covered without having to
+        # remember it. Idempotent, so the already-clean portion is unaffected.
+        message = (
+            sanitize_incoming_prompt(message)
+            if runtime_authored
+            else sanitize_untrusted_text(message)
+        )
 
     # Binary uploads can't be forwarded over the text-only ACP prompt channel.
     if files:
--- a/src/suzent/acp/runtime.py
+++ b/src/suzent/acp/runtime.py
@@ -170,6 +170,7 @@ async def stream_acp_turn(
     *,
     files: list[Any] | None = None,
     file_mentions: list[Any] | None = None,
+    runtime_authored: bool = False,
 ) -> AsyncGenerator[str, None]:
     db = get_database()
     chat = db.get_chat(chat_id)
@@ -189,6 +190,7 @@ async def stream_acp_turn(
     from suzent.core.system_reminder import (
         extract_system_reminder_display_trigger,
         sanitize_incoming_prompt,
+        sanitize_untrusted_text,
         strip_system_reminders,
     )
 
@@ -205,8 +207,20 @@ async def stream_acp_turn(
     # would lose text the user meant to send, and a message that was nothing but
     # a block would empty out and slip past the truthiness check below as a blank
     # turn.
+    #
+    # Provenance comes from *runtime_authored*, never from the token in the text.
+    # RUNTIME_NONCE is embedded in every reminder the model reads, so it is a
+    # bearer token the untrusted side can observe and replay: honouring it on
+    # externally supplied input would let a message that echoes it back be
+    # treated as runtime context and vanish from the transcript. Only the
+    # internal caller that built the reminder may claim that status, and it says
+    # so on the call rather than in the string.
     if message:
-        message = sanitize_incoming_prompt(message)
+        message = (
+            sanitize_incoming_prompt(message)
+            if runtime_authored
+            else sanitize_untrusted_text(message)
+        )
     user_message = message
 
     display_trigger = extract_system_reminder_display_trigger(user_message)
@@ -404,9 +418,13 @@ async def run_acp_turn_text(
     message: str,
     config_override: dict[str, Any] | None = None,
     stream_queue: Any | None = None,
+    *,
+    runtime_authored: bool = False,
 ) -> str:
     text = ""
-    async for chunk in stream_acp_turn(chat_id, message, config_override):
+    async for chunk in stream_acp_turn(
+        chat_id, message, config_override, runtime_authored=runtime_authored
+    ):
         if stream_queue is not None:
             await stream_queue.put(chunk)
         if chunk.startswith("data: "):
--- a/src/suzent/acp/runtime.py
+++ b/src/suzent/acp/runtime.py
@@ -186,11 +186,22 @@ async def stream_acp_turn(
     # Only the agent sees that annotation; the transcript keeps what the user
     # typed. Comparing the annotated text against the stored row defeated the
     # duplicate check below and persisted the message twice.
-    user_message = message
     from suzent.core.system_reminder import (
         extract_system_reminder_display_trigger,
+        sanitize_stored_user_prompt,
         strip_system_reminders,
     )
+
+    # Sanitize before deriving the transcript *and* before running the turn, so
+    # the two cannot disagree. Without this, a message wrapping its payload in a
+    # nonce-shaped reminder block strips to nothing visible and persists only its
+    # own chosen display trigger, while the raw text still reaches
+    # _stream_prompt() below — a prompt that executes but is misrepresented in
+    # the audit transcript. Runtime-authored blocks carry our token and survive,
+    # so genuine cron and heartbeat triggers still render as trigger rows.
+    if message:
+        message = sanitize_stored_user_prompt(message)
+    user_message = message
 
     display_trigger = extract_system_reminder_display_trigger(user_message)
     visible_user_message = strip_system_reminders(user_message)
--- a/src/suzent/acp/runtime.py
+++ b/src/suzent/acp/runtime.py
@@ -188,7 +188,7 @@ async def stream_acp_turn(
     # duplicate check below and persisted the message twice.
     from suzent.core.system_reminder import (
         extract_system_reminder_display_trigger,
-        sanitize_stored_user_prompt,
+        sanitize_incoming_prompt,
         strip_system_reminders,
     )
 
@@ -199,8 +199,14 @@ async def stream_acp_turn(
     # _stream_prompt() below — a prompt that executes but is misrepresented in
     # the audit transcript. Runtime-authored blocks carry our token and survive,
     # so genuine cron and heartbeat triggers still render as trigger rows.
+    #
+    # The ingress variant, not the history one: this message is being sent now,
+    # so forged delimiters are escaped in place rather than deleted. Dropping
+    # would lose text the user meant to send, and a message that was nothing but
+    # a block would empty out and slip past the truthiness check below as a blank
+    # turn.
     if message:
-        message = sanitize_stored_user_prompt(message)
+        message = sanitize_incoming_prompt(message)
     user_message = message
 
     display_trigger = extract_system_reminder_display_trigger(user_message)
--- a/src/suzent/agent_manager.py
+++ b/src/suzent/agent_manager.py
@@ -365,7 +365,14 @@ def create_agent(
     )
     from suzent.tools.capability import RegisteredToolCapability
 
+    from suzent.core.system_reminder import (
+        make_tool_output_sanitizer_history_processor,
+    )
+
     capabilities = [
+        # Sanitizer first: it must neutralize forged reminder delimiters before
+        # compaction folds tool output into a summary that we can no longer inspect.
+        ProcessHistory(make_tool_output_sanitizer_history_processor()),
         ProcessHistory(make_compaction_history_processor()),
         ToolSearch(),
     ]
--- a/src/suzent/core/agent_inbox.py
+++ b/src/suzent/core/agent_inbox.py
@@ -296,6 +296,9 @@ class AgentInboxDispatcher:
                     else delivered_content,
                     config_override,
                     None,
+                    # This text was wrapped by us a line ago; the flag is the
+                    # provenance, since the token inside it proves nothing.
+                    runtime_authored=is_subagent_result,
                 )
             else:
                 await ChatProcessor().process_background_turn(
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -782,7 +782,17 @@ class ChatProcessor:
                 )
 
         # --- System Reminder Injection (includes per-turn RAG hook when memory enabled) ---
-        from suzent.core.system_reminder import build_combined_reminder
+        from suzent.core.system_reminder import (
+            build_combined_reminder,
+            sanitize_untrusted_text,
+        )
+
+        # Neutralize reminder delimiters in everything we did not author, before
+        # the genuine reminder is appended below. Otherwise a user message or an
+        # attachment carrying its own delimiters would be indistinguishable from
+        # runtime context once concatenated.
+        message_content = sanitize_untrusted_text(message_content)
+        attachment_context = sanitize_untrusted_text(attachment_context)
 
         # Pass the raw user message so per-turn hooks (e.g. dynamic RAG retrieval)
         # are invoked. Heartbeats and pure tool-resume turns pass None so those
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -413,6 +413,20 @@ def _stripped_image_reminder(virtual_paths: list[str]) -> str | None:
     )
 
 
+def build_steering_text(steer_message: str) -> str:
+    """Render an interruption into the prompt text appended to history.
+
+    Steering re-enters ``process_turn`` with an empty ``message_content``, so the
+    ingress sanitizer there never sees this string, and the tool-output processor
+    deliberately skips ``UserPromptPart``. This is the only place it can be
+    sanitized, which is why the wrapping lives in one testable function rather
+    than inline at the call site.
+    """
+    from suzent.core.system_reminder import sanitize_untrusted_text
+
+    return f"[User interrupted to redirect]: {sanitize_untrusted_text(steer_message)}"
+
+
 class ChatProcessor:
     """Encapsulates the lifecycle of a single conversation turn."""
 
@@ -1536,7 +1550,7 @@ class ChatProcessor:
         # 4. Append steering message
         from pydantic_ai.messages import ModelRequest, UserPromptPart
 
-        steering_text = f"[User interrupted to redirect]: {steer_message}"
+        steering_text = build_steering_text(steer_message)
         message_history.append(
             ModelRequest(parts=[UserPromptPart(content=steering_text)])
         )
--- a/src/suzent/core/context_compressor.py
+++ b/src/suzent/core/context_compressor.py
@@ -722,6 +722,15 @@ class ContextCompressor:
                 )
             ]
         )
+        # The summary is model output derived from tool content an attacker may
+        # control, so it can be steered into emitting literal reminder
+        # delimiters — including by reassembling them from an escaped form. It is
+        # inserted after the sanitizing history processor has run and lands in
+        # persisted history, so it is neutralized here at the point of creation.
+        from suzent.core.system_reminder import sanitize_untrusted_text
+
+        summary = sanitize_untrusted_text(summary)
+
         summary_response = ModelResponse(
             parts=[
                 TextPart(
--- a/src/suzent/core/system_reminder.py
+++ b/src/suzent/core/system_reminder.py
@@ -378,6 +378,41 @@ _ANY_XML_TOKENED_BLOCK_RE = re.compile(
 )
 
 
+def _split_preserving_own_blocks(text: str, handle_span) -> str:
+    """Apply *handle_span* to everything except blocks bearing our own token.
+
+    Both callers need the same split — keep what this process authenticated,
+    treat the rest as untrusted — and differ only in what "treat" means.
+    """
+    if not text:
+        return text
+    out = []
+    last = 0
+    for match in _OWN_BLOCK_RE.finditer(text):
+        out.append(handle_span(text[last : match.start()]))
+        out.append(match.group(0))
+        last = match.end()
+    out.append(handle_span(text[last:]))
+    return "".join(out)
+
+
+def sanitize_incoming_prompt(text: str) -> str:
+    """Neutralize forged delimiters in a message arriving now, losing nothing.
+
+    The ingress counterpart to ``sanitize_stored_user_prompt``. The difference is
+    deliberate: stored history may contain machine-authored blocks from earlier
+    processes that are stale and safe to drop, but a message being sent right now
+    is the user's, and deleting part of it loses words they meant to send. Paste
+    a raw reminder block in to ask about it and you should see it echoed back
+    escaped, not silently swallowed — and if it was the whole message, not turned
+    into an empty turn.
+
+    Blocks carrying our token are preserved, so a runtime-authored cron or
+    heartbeat prompt still arrives intact.
+    """
+    return _split_preserving_own_blocks(text, sanitize_untrusted_text)
+
+
 def sanitize_stored_user_prompt(text: str) -> str:
     """Make a user prompt restored from history safe to send again.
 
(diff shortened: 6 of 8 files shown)


cyzus:
Both findings confirmed and fixed in `08d70006`. Thanks — the second one in particular was the exact failure mode this PR claims to close.

**Structured tool payloads (P1):** correct, and it defeated the protection where it mattered most. `webpage_tool._crawl_url` returns `ToolResult`, a Pydantic model with the fetched markdown in `.message`, so `isinstance(content, str)` skipped it outright. Added `sanitize_untrusted_payload()`, which walks strings, lists, tuples, dicts and model fields (depth-capped at 6) and returns a sanitized copy via `model_copy`, falling back to `setattr` for models that reject it. Clean payloads are returned identically, so the no-op path stays allocation-free.

**Tag spellings (P1):** also correct. Ingress used literal `.replace()` on the exact lowercase tags while `_ANY_XML_STRIP_RE` is `IGNORECASE` — so `<SYSTEM-REMINDER>` reached the model *and* was stripped from the transcript. Model sees it, user does not, which is the asymmetry the whole change is supposed to remove. Now case-insensitive regex tolerating attributes and inner whitespace (`< system-reminder >`, `<system-reminder foo='bar'>`).

I also neutralized `<system-reminder-display-trigger>` while in there — a forged one would surface attacker text in the UI as though the runtime had raised it. Same class of bug, adjacent tag.

9 new tests covering both, parametrized over the tag spellings. Suite 1656 passed, ruff clean.

Not merging until you have had a look at the follow-up.

cyzus:
@codex review

cyzus:
All three confirmed and fixed in `fede960f`. Each was a way around the previous round, which is a fair signal that the first pass was too shallow.

**Steering prompts (P1):** correct, and the most consequential. `process_steer` appends `steer_message` to history as a `UserPromptPart` and re-enters `process_turn` with `message_content=""` — so ingress never sees it, and the tool-output processor skips `UserPromptPart` by design. Nothing covered it. Extracted `build_steering_text()` so the wrapping and the sanitizing sit in one testable function rather than inline, and tested behaviourally rather than by asserting on source.

**Mapping keys (P1):** correct. I sanitized values and left keys raw, and a tool or MCP response chooses its own property names. Keys now take the same path.

**Depth cutoff (P1):** correct, and the sharpest of the three — the guard I added to prevent runaway recursion was itself an attacker-reachable door, since past six levels the value came back untouched. Replaced with identity-based cycle detection plus a much higher ceiling that *redacts* the branch rather than returning it. It now fails closed. Scalars short-circuit ahead of the cycle bookkeeping so the common path doesn't pay for it.

6 new tests: deep-nesting redaction driven off the constant, self-referential payloads, forged mapping keys, and steer text in both forged and ordinary form.

Suite 1661 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `a45e0c5`.

**Legacy user prompts (P1):** correct. Ingress only protects new messages; chats created before this change were never sanitized on the way in, and the processor skipped `UserPromptPart` outright — so a forged block saved back then stayed trusted by the model on every request *and* stayed hidden by `strip_system_reminders()`. That is precisely the asymmetry this PR claims to end, preserved indefinitely for existing chats.

`sanitize_stored_user_prompt()` keeps blocks bearing any well-formed runtime token and neutralizes everything else. Matching *any* token rather than only this process's matters: tokens change on restart, so a stricter check would neuter every genuine reminder after each deploy, not just legacy ones.

**⚠️ Behaviour change I want explicitly on the record before merge.** Pre-token history is indistinguishable from a forgery, so this fails closed: genuine reminders in chats predating this change lose their delimiters and become *visible text* in those transcripts. Display is rebuilt from the pydantic history (`_rebuild_display_messages`), so in-place neutralization does reach the UI — I checked whether returning copies could avoid it, but turn-end persistence reads the processor's output, so the copy would be persisted too.

One-off cohort, and I think security is the right direction here, but it is a visible change to existing user data rather than a pure internal fix. Happy to gate it behind a flag or a one-time migration instead if that is preferred.

4 new tests: forged legacy blocks neutralized, genuine current-process reminders surviving, previous-process tokens surviving, and forged text around a genuine block still cleaned. Suite 1665 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `dcdc620`. The second is the most serious finding in this PR so far, and it invalidates a claim I made in the description.

**Nonce-shaped blocks (P1):** correct. `_TOKENED_BLOCK_RE` matched *any* 16 hex characters, so `PUA_START + "0"*16 + payload + "0"*16 + PUA_END` was accepted as runtime-authored, extracted as a reminder, and stripped from the transcript. The mechanism I added to make blocks unforgeable was itself forgeable by anyone who noticed its shape. I introduced this in the previous commit while fixing the legacy-history finding, and my own comment on it described matching any token as a deliberate improvement — it was a hole.

The underlying constraint is that provenance cannot be verified across a restart: earlier processes' tokens are unknowable, so "was this ours?" is unanswerable for stored history. Rather than keep approximating it, unauthenticated blocks are simply not trusted, split by delimiter kind:

- **PUA blocks are dropped.** Invisible control characters are not typed by hand, so an unauthenticated one is either a forgery or an earlier process's reminder — hostile or stale. Dropping means the transcript does not change, because those blocks were already hidden.
- **XML tags are escaped, not dropped.** Someone can plausibly type `<system-reminder>` while discussing this feature, and deleting their words is worse than showing them.

**This retires the behaviour change I flagged on the previous commit.** Genuine pre-token reminders are now removed rather than made visible, so old transcripts render exactly as before and no flag or migration is needed. It also stops stale plan snapshots accumulating across restarts, which is a separate item in the audit doc.

**Multimodal prompts (P1):** correct. Image turns persist as `UserPromptPart(content=[full_prompt, *agent_images])`, and my string-only check skipped every one. Sequence content now has its string elements sanitized with media passed through untouched.

Also updated two tests from the previous commit that asserted the old defuse-in-place semantics for PUA — dropping is deliberate, so those assertions were wrong rather than the code.

7 tests here. Suite 1669 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `236952be`.

Correct. My escape-don't-delete rule for XML was justified as protecting text a human typed, but I applied it to *every* XML block including complete ones carrying `nonce="a1b2c3d4e5f6a7b8"` — which no one writes by hand. Under `SUZENT_XML_SYSTEM_REMINDER`, a restart leaves every stored reminder tagged with the previous process's token, so those got escaped rather than dropped: body stranded as text the display path can no longer strip, internal plan state surfacing as the user's own message, and stale context still resent to the model.

Dropping is now keyed on shapes people do not produce — PUA delimiters, and XML blocks with a nonce attribute — while bare `<system-reminder>` tags stay escaped so a human's words are never deleted. That keeps the reasoning behind the original rule and drops the over-application.

3 tests: stale tokenized XML dropped body-and-all, bare tags still escaped, and the current process's XML reminder surviving. Suite 1672 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `f9593a2a`.

Correct, and a good catch on a case I had no reason to be confident about: I reasoned about stale reminders purely as model-only context, and missed that a cron or heartbeat turn has *no user text at all* — the nested `display_trigger` is its entire visible record. Dropping the block wholesale emptied the prompt, and the next `_persist_state()` rebuild turned the former `system_triggered` row into a blank user message. Visible history corruption for any such chat reopened after a restart.

The trigger is now carried across when its block is dropped, with its inner text sanitized rather than preserved verbatim — the label survives, the model-only body still goes, and a forged block cannot smuggle delimiters back in through the fragment we keep. A dropped block with no trigger still leaves nothing behind.

3 tests: trigger survives while the body is dropped, nested delimiters inside a preserved trigger are neutralized, and triggerless stale blocks vanish cleanly. Suite 1675 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `e92174b3`.

Correct, and my previous fix simply did not work. I emitted the recovered trigger as a bare XML tag outside any reminder block, but `_rebuild_display_messages()` only looks for a trigger once `strip_system_reminders()` leaves nothing visible. A bare tag isn't inside a block, so `clean_text` stayed non-empty and the row persisted as an ordinary user message showing raw markup — and a later sanitizing pass would escape the tag into something unextractable.

Now rewrapped in a block carrying this process's token: strip clears it, the rebuild finds the trigger and emits `system_triggered` as before, and the token makes repeated passes idempotent, which matters because the processor runs before every model request.

The more useful correction is to how I tested it. Last round I asserted on `sanitize_stored_user_prompt()`'s output — `extract_system_reminder_display_trigger(out)` returned the right string, so the test passed while the actual rebuild was broken. I'd verified the fragment, not the outcome. This round drives `_rebuild_display_messages()` directly and asserts the row is `system_triggered`, which fails against the previous commit.

3 tests: end-to-end rebuild produces the trigger row, recovery is idempotent, model-only body still dropped. Suite green, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `fd83986`.

**Re-authenticating recovered triggers (P1):** correct, and the worst finding on this PR. Rewrapping the trigger in a block carrying `RUNTIME_NONCE` stamped our token onto content the same function had *just classified as unauthenticated* — so a forged pre-change prompt could have its trigger laundered into trusted hidden context the model is told to obey, and shown in the UI as a system trigger. I introduced that one round ago while fixing the bare-tag problem, which I introduced the round before while fixing the blank-row problem.

Recovered triggers are now emitted as an ordinary labelled line: the record of what fired survives, so a cron chat does not rebuild as an empty row, but it is visible text carrying no token. It no longer rebuilds as a `system_triggered` row — that is the price of not signing what cannot be verified, and the right price.

**Dataclass tool results (P1):** correct. pydantic-ai accepts and JSON-serializes dataclasses, and the walker treated only `model_fields` objects as structured, so a dataclass field holding fetched text reached the model with delimiters intact. Now traversed via `dataclasses.replace` with a `setattr` fallback for frozen instances.

---

**A recommendation on how to finish this PR.** Three consecutive rounds have been spent on trigger recovery, and each of my fixes introduced the next finding: blank rows → bare tag that does not rebuild → token laundering. That is not a run of bad luck, it is a sign I am patching a design that is wrong at the root. Recovering display metadata by parsing it back out of model-visible history means repeatedly deciding how much to trust text that by definition cannot be authenticated.

The right fix is to persist trigger metadata when the turn is created, so it never has to be recovered from history. That is a separate change touching persistence, and I have deliberately not attempted it here.

So I would suggest: **treat the trigger-recovery behaviour in this PR as a human review point rather than something to iterate on further.** The security posture is now fail-closed and I am confident in it; what I am not confident in is my own judgment about the display trade-offs, given three attempts. Everything else in the PR — ingress, tool output, stored history, tokens — has been stable for several rounds.

7 tests here. Suite 1682 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `6f0eb62f`.

Correct on all three shapes. But the types matter less than the reason they kept being missed, so I fixed that instead.

The walker listed the shapes it understood and returned everything else unchanged. That default is backwards for a sanitizer: every unlisted type is a hole, and they surfaced exactly one per review round — Pydantic model, then dataclass, now set/frozenset/Path. Patching in `set` and `Path` would have set up round ten with `deque`, `UUID`, `Decimal`, or an enum.

So the traversal is now **fail-closed**. Containers and structured objects are still walked, but an unrecognized value is rendered the way the serializer would render it and kept only if that text is clean — otherwise it degrades to sanitized text. Anything that cannot be rendered at all is redacted. A tool adding a new return type can no longer silently reopen this.

I also pulled the Pydantic and dataclass field-walking into one `_sanitize_fields` helper; they had already drifted in their fallback behaviour, which is its own small bug waiting to happen.

11 tests, including a parametrized sweep over container kinds and an `Exotic` object whose `__str__` carries delimiters — that last one fails against every previous version of this function. Suite 1693 passed, ruff clean.

The trigger-recovery recommendation from my previous comment still stands and is unaffected by this change.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `f6a9bbae`. Both were introduced by the previous commit, and the first one directly falsifies what I claimed about it.

**Unfixable objects passed through (P1):** correct. `_sanitize_fields` returned the original when `rebuild` and every `setattr` failed, shipping the delimiters it still contained. I described that commit as making the walker fail closed; on this path it did not, and I did not test it. It redacts now, and there is a test with an object that refuses both copy and `setattr`.

**NamedTuple returns (P2):** correct, and worse than a missed case — I *broke* a working shape. Switching `tuple(...)` to `type(value)(...)` for subclass preservation meant a NamedTuple got one generator where it expects one argument per field, raising inside the history processor and aborting the model request. Even for completely clean results. `_rebuild_sequence` now constructs per shape and degrades to the plain builtin when a subclass cannot be rebuilt.

---

I want to be straight about the pattern rather than just keep posting fixes.

Ten rounds, eighteen findings, no false positives. Eight of them are mine, created while fixing the round before. Last round I said the fail-closed inversion was structural and I expected that class to stop — one round later, two findings in that exact code, one of them a fail-open hole in the inversion itself and one a regression in a shape that previously worked.

The security direction is right and each individual fix here is, I believe, correct. But my confidence in my own convergence has been wrong every time I have expressed it, so I am not going to express it again.

**Concretely, what I would ask for:** a human pass over `sanitize_untrusted_payload` and the trigger-recovery path before this merges. Not because any specific defect is outstanding, but because the evidence says my error rate in this file is high and Codex is finding what I miss one layer at a time rather than all at once. If it would help, I can split the payload walker into its own PR so the stable parts — ingress, tool-output routing, stored-history handling, tokens — can land separately.

5 tests here. Suite 1698 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `9a159c0e`.

**Enum values (P1):** correct, and it exposes a flaw in the fail-closed check itself rather than another missing type. I rendered with `str()` and treated that as what the model would see — but `str(Payload.BAD)` is `'Payload.BAD'`, clean, while pydantic serializes the member's *value*. So the delimiters sailed through the very branch added to stop unknown shapes. Now rendered via `pydantic_core.to_json(value, fallback=str)`, which is the representation the tool-return serializer actually emits. Checking anything other than the wire form was the mistake.

**Token in debug logs (P2):** correct, and it invalidates a claim in this PR's description. I wrote that the token 'cannot be produced by anything that cannot read this process's memory' — but `build_combined_reminder()` logged the fully wrapped result at DEBUG and file logging records DEBUG unconditionally, so every reminder wrote `RUNTIME_NONCE` to disk. Anyone who can read a log file could forge a block. The log is metadata only now (chat, part count, length), which also stops it spilling retrieved memories, goals and local paths. Thanks for the `AGENTS.md:35` pointer — 'never log secrets' is exactly the rule I broke.

Note this only fixes the reminder-side log. The full-system-prompt debug log at `chat_processor.py:976` has the same problem for core memory and repo context; it is tracked in the audit doc and I have deliberately left it out of this PR rather than widen the scope again.

3 tests: an enum whose value carries delimiters, a clean enum passing through untouched, and a log assertion that the token and the reminder body stay out of `caplog` while the returned reminder is still tokenized. Suite 1701 passed, ruff clean.

The request from my previous comment stands unchanged.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `d38f7ba4`.

Correct. A computed field, serialization alias, or custom `model_serializer` all introduce content that `model_fields` cannot see, so finding the raw attributes clean proves nothing about what `model_response_str()` emits.

I did not add handling for those three. This is the fifth round of findings on this one function — unlisted types, unfixable objects, enum values, serializer hooks — and they are all the same mistake: I keep validating **structure** when the thing that reaches the model is a **serialization**. A pre-condition has to enumerate what it handles; the list will never be complete.

So the invariant is now a post-condition. `sanitize_tool_payload()` runs the structural walk, then renders the result the way the tool-return serializer will and replaces it with sanitized text if any delimiter survived. The walk is demoted to best-effort structure preservation — nice for keeping a `ToolResult` a `ToolResult` — and correctness no longer depends on it being complete. One check, every shape, including shapes nobody has thought of yet.

That is what this fix should have looked like from the start, and I would rather say so than present it as an incremental improvement.

4 tests, one per serializer hook plus a clean model passing through by identity. Suite 1705 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `1d83eab9`.

Correct, and it is a gap of a different kind from the last several: not a shape I mishandled, but **an entry path I never looked at**. `stream_acp_turn()` derived the persisted row from a stripped copy while passing the raw text to `_stream_prompt()`, so the transcript and the executed prompt came from different strings. A turn wrapping its payload in a nonce-shaped block stripped to nothing visible, persisted its own chosen label as a `system_triggered` row, and still ran the hidden instruction.

Fixed by sanitizing once up front and deriving both from that text. `sanitize_stored_user_prompt` preserves blocks carrying our token, so genuine cron and heartbeat turns still render as trigger rows.

Worth noting what this means for the PR as a whole: I treated the chat processor as *the* ingress and built everything around it. ACP is a second door into the same trust boundary and my ingress fix simply did not cover it. That is a scoping error rather than a coding one, and it makes me less confident that the remaining entry points have all been enumerated — the social path and any other caller reaching `_stream_prompt` or `process_turn` directly deserve the same look, which I have not done.

3 tests, written as a before/after on the actual display derivation: the forged block persists as `system_triggered` with a benign label before sanitizing and as an ordinary user row with the payload gone after, plus a genuine cron trigger still rendering correctly and an ordinary message untouched. Suite 1708 passed, ruff clean.

My earlier request stands, and I would now widen it: the entry-point enumeration wants a human eye as much as the payload walker does.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `991369e9`.

Correct, and self-inflicted: I reused `sanitize_stored_user_prompt` at an ingress point one commit after writing a docstring explaining that its whole purpose is dropping stale history. The drop semantics are right for restored history and wrong for a message being sent now — pasting a raw block in to ask about it made it vanish, and a message that was *only* a block emptied out and slipped past the truthiness check as a blank turn.

Now split by intent: `sanitize_incoming_prompt` escapes in place and keeps every character; `sanitize_stored_user_prompt` keeps deleting for history. Both preserve our own token so cron and heartbeat turns still render as trigger rows, and the shared split lives in one helper so they cannot drift apart.

The ACP tests now exercise the function ACP actually calls — the previous round's tests asserted against `sanitize_stored_user_prompt` and would have kept passing through this bug, which is the same verify-the-fragment mistake I made earlier with the display trigger. Added a test pinning the difference between the two sanitizers, since confusing them is precisely what happened.

5 tests. Suite 1710 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `5fde672`.

**Replayable token (P1).** This is the important one and it invalidates the central premise of the PR as I originally described it. `RUNTIME_NONCE` is embedded in every reminder the model reads — it is a bearer credential visible to the untrusted side. Honouring it on externally supplied input meant a message that echoed the token back was treated as runtime context: stripped from the transcript or recorded under its own chosen display trigger, while its payload still ran. The audit-spoofing hole, reopened through the front door by the very mechanism meant to close it.

A secret you show to the party you are defending against is not a secret. I built the whole design around that mistake and did not notice it across fourteen rounds; you did.

Provenance now comes from the call path. `stream_acp_turn` / `run_acp_turn_text` take `runtime_authored`, and only `agent_inbox` — one line after wrapping the reminder itself — sets it. Everything else is sanitized outright, token or no token. The history check still uses the token, and that remains sound *because* ingress no longer honours it: the only way an own-token block reaches a stored prompt is if we put it there. That is a real argument rather than an assumption, and it is the first time in this PR the token has had one.

**Pre-write reconciliation (P2).** Correct. `/chat/send` and ACP-steer pre-wrote the display row from the raw message before the turn sanitized it, so the reconciliation compared different strings and appended a second escaped copy beside the raw one. Both routes now sanitize before pre-writing; the turn's pass is a no-op because escaping is idempotent, which a test pins rather than assumes.

3 tests: a replayed token buying no trust, the declared internal path keeping its trigger row, and idempotency.

Suite 1713 passed, ruff clean.

---

I am going to stop asking for review politely and just say it plainly: **this PR should not merge on my judgement.** Fifteen rounds, twenty-five findings, none of them wrong. The last one shows the core mechanism was unsound from the first commit in a way I could not see by iterating. The current state is, I believe, correct — but I believed that at rounds 9, 10 and 12 too, and I was wrong each time.

What I would suggest concretely: land the parts that have been stable for many rounds — ingress sanitizing, tool-output routing, stored-history handling — and treat the token mechanism and trigger recovery as a separate change with a human designer on it. I am happy to do that split on request. I will keep responding to review, but I do not think more rounds from me is the way this gets finished.

cyzus:
@codex review

cyzus:
All three confirmed and fixed in `97765a9a`.

**File annotations (P1):** correct. `_build_acp_file_context()` interpolates caller-supplied paths and is prepended *after* the message is sanitized, so a forged path reached the model while the transcript showed only the ordinary user text. Now the assembled string is sanitized rather than only its parts — which is the same post-condition idea as `sanitize_tool_payload`, applied at ingress, and means anything later added at that point is covered without someone having to remember it.

**Mapping key collisions (P2):** correct, and a data-corruption bug I introduced while fixing a security one. A forged key and its already-escaped twin collapse to the same string and one entry vanishes. Falls back to sanitized text retaining every entry.

**Multimodal tuple subclasses (P2):** correct, and the least excusable of the three. This is the *same* NamedTuple defect you flagged six rounds ago, at a second site. I wrote `_rebuild_sequence` specifically for it and then did not apply it here. I fixed the instance instead of the pattern; both sites share the helper now.

That last one is the clearest single illustration of the failure mode in this PR. My fixes are local to whatever was pointed at, and the same defect elsewhere survives untouched until someone points at that too.

3 tests. Suite 1716 passed, ruff clean.

My recommendation from the previous comment is unchanged and I would rather it were acted on than that I keep going: split the stable parts out and put a human on the token mechanism and trigger recovery. Sixteen rounds, twenty-eight findings, none of them wrong.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `9b622ca0`.

Correct, and this is the same defect I fixed for mapping keys in the previous commit — at the branch directly above it. Third time in this PR I have fixed an instance and left its twin: NamedTuple reconstruction (rounds 10 and 16), and now collision handling (rounds 16 and 17).

So this time I did not patch the second site. Both call `_degrade_if_collapsed`, and I audited the remaining branches instead of waiting to be told about them: lists and tuples preserve duplicates, named fields are keyed by name, so the deduplicating containers are the only two that can lose data this way. There is a test asserting ordered containers keep their shape rather than taking the text fallback, so the guard cannot quietly spread to cases that do not need it.

I am aware that 'this time I checked the neighbours' is what I should have been doing for the last several rounds, and that saying so does not make the previous rounds better.

4 tests. Suite 1720 passed, ruff clean.

Seventeen rounds, twenty-nine findings, none wrong. My recommendation is unchanged: split the stable parts out, put a human on the token mechanism and trigger recovery. I will keep fixing what review finds, but the case for finishing this PR by iteration got weaker again this round, not stronger.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `d1c284d2`. Both are P1s in the security surface, so my note last round that it had 'held for two rounds' was wrong — I keep reading trends into this and being wrong, and I should stop doing that.

**Compaction summary (P1):** correct, and it defeats something I explicitly designed for. My first commit registered the sanitizer *before* the compactor with a comment saying it was ordered that way 'so forged delimiters cannot slip through folded into a summary'. That reasoning was wrong: the summary does not exist when that pass runs. It is generated by a model from attacker-influenceable tool content, can be steered into emitting delimiters — including by reassembling them from an escaped form — and is inserted afterwards as a synthetic `ModelResponse`, which the processor skips. Now neutralized where it is created.

**Upload paths in the hidden image reminder (P1):** correct, and the general form is worse than the reported case. Wrapping is what confers trust, and *nothing* going into a block was being checked — not just AG-UI paths, but every hook fragment: retrieved memories, goal and task text, background agent results. All user-influenced, all wrapped and authenticated. So the fix is at the wrap point, not at the one caller you found; a new reminder source cannot forget it now.

That second one is the fourth time a finding has turned out to be one instance of a class. I have started checking siblings before replying, which is how the wrap-point fix came about rather than a patch to `_stripped_image_reminder`.

4 tests. Suite 1724 passed, ruff clean.

Eighteen rounds, thirty-one findings, none wrong. Recommendation unchanged.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `aab5f94b`.

Correct, and it is the direct sibling of the compaction fix from the previous round: I sanitized *compressor-generated* model output and left *ordinary* model output alone. Adversarial content can steer the model into emitting delimiters in a `TextPart` or `ThinkingPart` — or reassembling them from escaped tool content — and that response is persisted and replayed as context.

Fifth time a finding has turned out to be one instance of a class, so I fixed the class. The processor handles `UserPromptPart` and tool parts explicitly and now sanitizes **any** other part carrying string content as a catch-all, so a part type added later is covered by default rather than becoming round twenty-something. I also included `ToolCallPart.args` on the same reasoning — model-authored and replayed — since it was exposed by the identical argument and had not been reported yet.

Tests include one asserting the genuine current-turn reminder still survives the catch-all, which is the obvious way this fix could have gone wrong.

5 tests. Suite 1729 passed, ruff clean.

Nineteen rounds, thirty-two findings, none wrong. Recommendation unchanged: split the long-stable parts out and put a human on the token mechanism and trigger recovery.

cyzus:
@codex review

PR #163, opened 2026-09-02
fix(memory): resolve core memory per request instead of per agent build

cyzus:
## Problem

Agents are cached and reused across chats and users. Core memory was captured in a closure at construction:

```python
@agent.instructions
def inject_memory_context(_: Any) -> str:
    return memory_context or ""
```

The `_` is the tell — the run context was never consulted. Editing `persona.md`, `user.md`, `MEMORY.md`, or a project's `context.md` had no effect until something unrelated forced the agent to rebuild.

The same closure is why the legacy LanceDB path was a cross-user hazard. `_user_id` is excluded from the agent cache key, and on that path `get_core_memory()` reads persona/user/facts *by* `user_id` — so two users sharing a model, tool set and project shared one snapshot. In practice `lifecycle.py` always supplies a markdown store, so the default deployment doesn't reach it; treating it as defence in depth.

## What this is not

Not an isolation fix. I want to be precise, because an earlier draft of the audit claimed cross-chat `context.md` bleed and that claim was wrong:

- `context.md` is **project**-scoped, via `get_project_dir(chat_id)`, not chat-scoped.
- `_project_context_dir` / `_working_context_dir` / `_repository_root` are already part of the stable config, so different projects already rebuild the agent.

The audit doc has been corrected. This PR is about staleness.

## Approach

- `inject_memory_context` is now async and resolves from `ctx.deps`. It falls back to the construction-time snapshot only when memory is disabled (`deps.memory_manager` is `None` — already exactly that condition) or the lookup raises.
- `MemoryManager.get_core_memory_context()` caches on a file-mtime revision, so a per-request read costs four `stat` calls rather than four file reads plus a re-render. The legacy path has no cheap revision and stays uncached — which is also what makes it correct.
- `_TRANSIENT_KEYS` moved to module scope so the cache-key invariant is testable instead of living in a comment. A test now fails if anyone adds `_project_context_dir` to it.

## Caching note

Core memory sits in the system prompt, i.e. inside the prompt-cache prefix. Resolving per request means an edit to a memory file invalidates that chat's prefix — correct, and rare relative to turn count. Chats already have distinct prefixes, so nothing is lost across chats. The alternative (moving core memory into the user message to protect the prefix) would re-bill it every turn and is explicitly not what this does.

## Tests

`tests/memory/test_core_memory_freshness.py` (7) — revision fingerprint behaviour including missing files, refresh after edit, cache hit on no-change, and the legacy path staying live per user.

`tests/prompts/test_core_memory_is_request_scoped.py` (5) — the injector reads the serving chat rather than the snapshot, returns empty when memory is off, falls back on error, and the cache-key guards.

Full suite: 1649 passed. `ruff check` / `ruff format` clean.

## Context

P1-0 in `docs/03-developing/system-prompt-and-reminder-audit.md`. Follows #162 (P0, forgeable reminder boundaries). Independent of it — no shared files.

### DIFF
--- a/src/suzent/agent_manager.py
+++ b/src/suzent/agent_manager.py
@@ -31,9 +31,6 @@ from suzent.prompts import (
 from suzent.skills import get_skill_manager
 
 # Import memory lifecycle functions (for backward compatibility re-exports)
-from suzent.memory.lifecycle import (
-    get_memory_manager,
-)
 
 # Suppress LiteLLM's verbose logging
 os.environ["LITELLM_LOG"] = "ERROR"
@@ -228,9 +225,7 @@ def _is_transient_stdio_error(message: str) -> bool:
     )
 
 
-def create_agent(
-    config: Dict[str, Any], memory_context: Optional[str] = None
-) -> Agent[AgentDeps, str]:
+def create_agent(config: Dict[str, Any]) -> Agent[AgentDeps, str]:
     """
     Creates a pydantic-ai Agent based on the provided configuration.
 
@@ -404,7 +399,6 @@ def create_agent(
     register_dynamic_instructions(
         agent,
         base_instructions=base_instructions,
-        memory_context=memory_context,
         session_guidance_items=session_guidance_items,
         enabled_model_ids=enabled_models,
         current_model_id=model_id,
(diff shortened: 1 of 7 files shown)


cyzus:
Both findings confirmed and fixed in `e077dc0d`.

**Resolver paths in the cache key (P2):** correct. Host mode renders cwd and the notebook mount into the section, so the same memory files produce different text for different resolvers — that makes the paths part of the section's identity, not incidental formatting. Rather than duplicate the derivation in the key, I extracted `_resolver_paths()` and share it between the renderer and the key, so the two can't drift apart later.

**Caching a failed render (P2):** correct, and I'd rate the impact above P2. `format_core_memory_for_context()` catches read errors and returns `""`; caching that under an unchanged revision meant one transient failure left the chat with no core memory until some unrelated file happened to change — a self-inflicted outage with no path back. Empty renders are no longer cached.

One adjustment while adding tests: I'd written a test asserting the rendered output differs when the notebook mount changes, and it failed — the mount is resolver state the section is built from, but the current template doesn't print it. Rather than delete the case or weaken the assertion, I test `_resolver_paths()` directly, so the mount stays in the cache identity and the test says why.

Suite 1653 passed, ruff clean. Holding off on merge pending your look.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `cd310114`.

**Unbounded cache (P2):** correct — and my own comment claiming it was "bounded by the number of live chats" was simply wrong, which is worth naming since that comment is what would have stopped the next reader from checking. Nothing evicts on chat deletion, and adding resolver paths to the key in the previous commit *made it worse*: every cwd or notebook-mount change now strands its predecessor permanently. Each stranded entry holds a full rendered prompt.

Swapped to an `OrderedDict` capped at `CORE_MEMORY_CACHE_MAXSIZE` (256) with `move_to_end` on both hit and write, so eviction is genuinely least-recently-used rather than insertion-ordered. Two tests: the ceiling holds under 40 distinct chats, and the LRU ordering evicts the coldest entry rather than the oldest-inserted.

Suite 1655 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `c749106b`.

**Snapshot fallback (P2):** correct, and I'd rate it higher — the fallback I added reintroduced the exact bug class the PR exists to remove. The agent is cached across chats and users, so the construction-time snapshot belongs to whichever request built it; on the legacy LanceDB path, where blocks are read by `user_id`, one transient failure hands A's persona and facts to B.

Rather than patch the fallback I removed the mechanism. With the error path returning `""`, the snapshot had no reader left, so `create_agent` and `register_dynamic_instructions` no longer take `memory_context` and `get_or_create_agent` no longer prefetches it — which also drops an `await` and a `PathResolver` construction from agent creation. Core memory is resolved per run from `ctx.deps`; on failure the section is omitted, because a prompt with no core memory beats one carrying someone else's.

Suite 1655 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `2aabd12`. This one is a regression introduced by this PR, not a pre-existing bug — worth stating plainly.

The old prefetch built its own `PathResolver(chat_id=..., sandbox_enabled=..., custom_volumes=...)` with **no cwd**, so `get_working_dir()` returned `project_dir` and the rendered `context.md` path coincided with where `read_session_context()` reads. Switching to `deps.path_resolver` — the whole point of the PR — brought an authorized `cwd` along with it, so a chat pointed at an external folder was told to update `<cwd>/context.md` while reads still resolved to the project directory. The agent's context updates would land in a file nothing ever reads back, and silently: no error, just memory that never persists.

Fixed by rendering `project_dir / "context.md"` explicitly rather than relying on the two paths coinciding. `PathResolver.project_dir` is `sandbox_data_path / "projects" / slug`, which is the same directory `_context_path(chat_id)` resolves to, so read and write now agree by construction for both resolver kinds.

Test asserts the rendered path follows the project directory and specifically *not* an authorized cwd. Suite 1656 passed, ruff clean.

cyzus:
@codex review

PR #167, opened 2026-09-02
fix(logging): trace prompt structure instead of dumping its contents

### LINKED ISSUE #165: Full system prompt is written to debug logs (core memory, repo context)
Follow-up from #162, and item P1-1 in `docs/03-developing/system-prompt-and-reminder-audit.md`.

`chat_processor.py` resolves and logs the **entire** system prompt at DEBUG — persona, `user.md`, `MEMORY.md`, the project `context.md`, and repository context. File logging records DEBUG unconditionally, so this lands on disk.

#162 fixed the equivalent leak on the reminder side (`build_combined_reminder` was writing `RUNTIME_NONCE` and the full reminder body to the log; it is metadata-only now) but deliberately left this one alone rather than widening scope again.

`AGENTS.md:35` — never log secrets or PII.

## Suggested fix

Metadata only by default: `chat=<id> sections=<ids> chars=<n> sha=<prefix>`. Put full-text prompt tracing behind a dedicated opt-in flag, not the DEBUG level, and redact known-sensitive sections even then.

cyzus:
Closes #165.

## Problem

`chat_processor` resolved and logged the **entire** system prompt at DEBUG: persona, `user.md`, `MEMORY.md`, the project `context.md`, repository instructions. File logging records DEBUG unconditionally, so all of that hit disk on every turn.

`AGENTS.md:35` — never log secrets or PII.

#162 fixed the equivalent leak on the reminder side (it was writing `RUNTIME_NONCE` and the full reminder body) but deliberately left this alone rather than widening scope mid-review.

## Approach

Prompt resolution is split into `resolve_system_prompt_sections()` returning `(name, text)` pairs, with `resolve_full_system_prompt()` built on it — so the sections stay separate and can be *described* without being printed.

The default line is now:

```
[SystemPrompt] chat=<id> chars=1234 sha=a1b2c3d4 sections=static:812, inject_date_context:56, inject_memory_context:9001, ...
```

This is more useful than the old dump for what the log is actually for — seeing which sections are present and how much of the budget each takes — and it carries no content. The `sha` gives you a stable identity to compare turns without revealing anything.

Full text now requires `SUZENT_PROMPT_TRACE`. Enabling debug logging no longer implies dumping user data; you have to ask for it.

## Note on the gating

The block stays behind the DEBUG check because resolution runs every instruction runner and is not free. The trace is opt-in on top of that.

## Tests

`tests/prompts/test_prompt_trace_redaction.py` — sections carry names, the joined sections still equal the full prompt (so the refactor is behaviour-preserving), and a trace line built the way the processor builds it contains the section sizes but none of their text.

Suite 1751 passed, ruff clean.

### DIFF
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -1004,8 +1004,12 @@ class ChatProcessor:
         # DEBUG unconditionally, so the previous full-text line put all of that
         # in the log on every single turn (AGENTS.md: never log secrets or PII).
         #
-        # Full text requires SUZENT_PROMPT_TRACE, a deliberate opt-in rather than
-        # a side effect of turning on debug logging.
+        # There is deliberately no flag to print the text. AGENTS.md states the
+        # prohibition without qualification, and the prompt carries core memory,
+        # permission feedback, sender names and project instructions — an opt-in
+        # does not make writing that to disk acceptable, it just moves who is
+        # responsible. The sha is enough to tell two turns apart, and the section
+        # sizes are what the log was actually useful for.
         try:
             if logger._core.min_level <= 10:  # DEBUG
                 import asyncio as _asyncio
@@ -1030,12 +1034,6 @@ class ChatProcessor:
                     _digest,
                     ", ".join(f"{name}:{len(text)}" for name, text in _sections),
                 )
-                if os.environ.get("SUZENT_PROMPT_TRACE"):
-                    logger.debug(
-                        "[SystemPrompt] full text for chat {} (SUZENT_PROMPT_TRACE):\n{}",
-                        chat_id,
-                        _prompt,
-                    )
         except Exception as e:
             logger.debug(f"[SystemPrompt] Failed to resolve debug prompt: {e}")
 
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -997,15 +997,23 @@ class ChatProcessor:
                 message_history, _budget
             )
 
-        # Debug-only: log complete system prompt (static + dynamic sections)
-        # as resolved by pydantic-ai instruction runners for this run context.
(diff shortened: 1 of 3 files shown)


cyzus:
@codex review

cyzus:
Confirmed and fixed in `776c1386`. Removed the escape hatch rather than repairing it.

Two things wrong, and the second is the one that matters. The lookup was a bare truthiness check, so `SUZENT_PROMPT_TRACE=false` or `=0` would have *enabled* the dump — a classic. But `AGENTS.md:35` states the prohibition without qualification, and you are right that an opt-in does not satisfy it. It just moves who is responsible for core memory, permission feedback, sender names and project instructions landing on disk.

So there is no flag now. The structural line stays: the sha distinguishes turns and the section sizes are what this log was actually useful for, and neither needs the text. That also makes the PR strictly smaller.

Added a test asserting no environment flag can turn it back on, so the hatch cannot be reintroduced quietly.

Suite 1752 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `db2544d0`.

Correct. I reached for a hash as a neutral-looking identifier without thinking about what it is over. A truncated unsalted digest of the prompt is a *verifier* for its contents: with the other sections known or predictable — a social run with memory disabled — someone holding the logs can hash candidate sender names or permission feedback until one matches. It also links identical private prompts across runs.

Removed. `chat_id` already correlates lines within a conversation, so it was buying nothing the trace needed. Section names and sizes remain, which is what the log was for.

Worth noting this is the second round on this PR where my instinct for 'safe metadata' was wrong in the same direction — first an opt-in flag for the full text, now a digest of it. Both looked like reasonable engineering conveniences and both leak the thing the change exists to stop leaking.

Test asserts no digest is computed in that block. Suite 1753 passed, ruff clean.

cyzus:
@codex review

PR #168, opened 2026-09-02
fix(display): restore cron trigger rows from the stored log after a restart

### LINKED ISSUE #164: Persist system-reminder display-trigger metadata instead of recovering it from history
Follow-up from #162. Not a known-exploitable defect — a design weakness that cost three review rounds and still has a residual behaviour cost.

## Problem

A cron or heartbeat turn has no user text: its entire visible record is a `display_trigger` nested *inside* the reminder block, in model-visible history. On restart the block can no longer be authenticated (see below), so it is dropped — and the trigger must be reconstructed by parsing it back out of text we have just decided we cannot trust.

Three attempts inside #162, each of which introduced the next bug:

1. Drop the block wholesale → cron chats rebuilt as blank user rows.
2. Re-emit the trigger as a bare tag → `_rebuild_display_messages()` only looks for a trigger once `strip_system_reminders()` leaves nothing visible, so it persisted as an ordinary user row showing raw markup.
3. Rewrap it in a block carrying `RUNTIME_NONCE` → **laundered attacker-authored triggers into trusted hidden context.**

Current behaviour is the safe one: recovered triggers are emitted as a plain labelled line. Correct, but a restored cron chat no longer renders a `system_triggered` row.

## Suggested fix

Persist the trigger as its own column/field on the message row when the turn is created, so it never has to be recovered from model-visible history. Display reads it directly; the reminder body stays model-only and can be dropped freely.

## Why it keeps biting

Display metadata and model context are two different things sharing one string. Every fix has to re-derive trust for text that by construction has none.

cyzus:
Closes #164.

## Problem

A cron or heartbeat turn carries no user text. Its entire visible record is the `display_trigger` nested *inside* the reminder block, in model-visible history.

The rebuild recovers it by parsing it back out of that history — which stops working the moment the block is dropped on restart. And dropping is correct: #162 established that text we cannot authenticate must not be trusted, and an unauthenticated block is either stale runtime output or a forgery. So after a restart, those turns rebuilt as an ordinary user row showing a plain `[system trigger: ...]` line.

This cost three rounds of review in #162, each attempt introducing the next bug — blank rows, then a bare tag that didn't rebuild, then a rewrap that **laundered attacker-authored triggers into trusted context**. That last one is why the current behaviour is deliberately the safe-but-lossy one.

## Approach

The row was correct when it was first written, and it is already in the stored log. So stop re-deriving trust for the text and take the answer from the log.

`_preserve_display_triggers` reinstates `system_triggered` rows exactly the way `_preserve_citation_sources` and `_preserve_permission_metadata` already reinstate metadata the rebuild cannot reconstruct — an established pattern in this file rather than a new mechanism, and no schema change.

It only overwrites the placeholder the drop leaves behind. A genuine user message sharing that index is left alone, which matters because positional matching is the weak part of this approach.

## Why this closes the issue rather than working around it

The issue asked for the trigger to be persisted so it never has to be recovered from model-visible history. That is what this does: display metadata now comes from the display log, and the reminder body stays model-only and freely droppable. The two concerns stop sharing one string.

## Tests

7 cases in `tests/core/test_display_trigger_preservation.py`: restoration over the placeholder, leaving a genuine user message alone, leaving an already-correct row alone, preserving sibling fields like timestamps, diverged history lengths, no stored log, and non-dict rows.

Suite 1755 passed, ruff clean.


cyzus:
@codex review

cyzus:
Confirmed and fixed in `57c349ae`.

Correct, and it invalidates the assumption I wrote into the docstring: *'the rebuild emits exactly one row per stored user turn'*. `_coalesce_unanswered_cron_triggers` breaks it directly — consecutive unanswered triggers collapse to one stored row while agent history keeps both turns. Pairing indices then restored the newest content over the older placeholder and stranded the real newest one as a user message, which the coalescing pass running afterwards could not clean up.

Matching is now on the label carried in the placeholder, so position stops mattering entirely — which also makes it robust to any other rebuild/stored divergence, not just this one.

The stored log still supplies provenance: a placeholder is reinstated only if that exact label was recorded as a trigger when the turn actually ran. That matters because the placeholder text is attacker-reachable — a forged block dropped from history leaves one too — so promoting placeholders unconditionally would let a forgery style itself as a system row, which is the laundering bug from #162. There is a test for that.

Duplicates are all restored rather than deduplicated here, because collapsing them is the coalescing pass's job and it runs immediately after; the test asserts the two passes compose.

4 new tests including the coalescing scenario end-to-end. Suite 1758 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `f2ca5a37`.

Correct, and the consequence you point out is worse than the misattribution: if the user's message lands next to an unanswered cron with the same label, the coalescing pass collapses them and **the user's turn disappears**. A display bug that eats input is not a display bug.

The root error was mine in the previous round. I replaced positional matching with label matching and treated 'this label was a real trigger once' as provenance for *this* row. It is not — it is provenance for some other row.

Recovered placeholders now carry `TRIGGER_MARK`, a PUA codepoint only the history sanitizer emits, which `sanitize_untrusted_text` strips from anything arriving from outside. So a caller cannot supply it. The stored-label check stays alongside: the mark proves the runtime wrote this line, the label proves that trigger really fired, and they cover different attacks — the mark alone would still let a *forged* dropped block style itself as a system row.

One wrinkle worth flagging since it is the kind of thing that breaks quietly: the processor runs before every model request, so a second pass would have stripped the mark off placeholders from the first and silently demoted them. `_scrub_untrusted_span` now preserves its own output, and idempotency is asserted rather than assumed.

2 new tests: an unmarked lookalike from a user stays a user row, and a mark supplied from outside is stripped before it can help.

Suite 1760 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `de4244a8`.

Correct, and it dismantles the claim I made last round. I wrote that the mark and the stored label were 'two independent things' covering 'different attacks'. They are not independent for legacy chats: the pre-sanitization rebuild parsed reminder blocks without authenticating them, so a forged trigger can already sit in the display log — and `_scrub_untrusted_span` stamps `TRIGGER_MARK` onto whatever block it drops, including that same unauthenticated one. Both pieces of evidence trace back to one forgery. Taking a forgery's own output as proof of its legitimacy is the laundering pattern this work has now hit three times, and I reintroduced it while arguing I had closed it.

Provenance is now recorded where it still exists rather than reconstructed after the fact. The display rebuild stamps `trigger_origin=runtime` on a trigger row only when the block it came from carried this process's token — something a forgery cannot produce — and restoration requires that stamp.

Legacy rows carry no stamp and are not restored, so those chats keep the plain text line. That is the honest outcome for history whose provenance is genuinely unrecoverable, and it is also what #164 actually asked for: metadata persisted when the turn is created, instead of re-derived from model-visible text.

2 new tests: a legacy unstamped row is not restored even with a marked placeholder, and the stamp is written for an authenticated block but withheld from a nonce-shaped forged one.

Suite passing, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `2ad095c2`.

Correct, and an obvious one once stated: the restored row *is* what the next save persists, so dropping the stamp meant the trigger survived one post-restart save and degraded on the following turn. A fix that works exactly once.

Carrying it forward vouches for nothing new — the chain still roots in the stamp written when the turn actually ran and the block carried this process's token.

I added a second test for the risk that creates: an unstamped legacy row must not be able to acquire a stamp by going round the loop. It cannot, because the stamp is only written where one already qualified. That is the failure mode I would expect from this change, and the previous three rounds on this PR were all variations of trust leaking across a boundary, so it seemed worth pinning rather than reasoning about.

2 tests, one walking two save cycles. Suite 1764 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `da3756b4`.

Correct, and it is the same class as your round-2 finding: I kept trying to establish per-occurrence provenance by **matching content**, and the label is precisely the part an attacker controls. `trigger_origin` narrowed which rows could contribute a label but never tied the proof to a turn. Five rounds, four of them variations on that one mistake.

So this revision stops matching content. Rows are matched on `timestamp`, which pydantic-ai sets when the part is created and which survives in serialized state — a real per-turn identity. A forged block in a different turn cannot borrow a genuine row's proof, because it cannot occupy that turn.

Three conditions, none substituting for another: the stored row is stamped `runtime` (a legacy log can hold a forged trigger row); the timestamps agree (ties the proof to this occurrence); the rebuilt row still looks like a placeholder this runtime emitted (a user cannot type one).

**One behaviour change worth naming:** a turn whose stored row was coalesced away now keeps the plain text line instead of being restored. I previously restored those and let coalescing collapse them — but nothing vouches for that turn, and manufacturing proof is how the last four attempts went wrong. Test asserts it explicitly rather than leaving it implied.

Test file rewritten around identity rather than patched: 16 cases including a shared label across different turns, legacy unstamped rows, unmarked user lookalikes, marks supplied from outside, coalesced turns, two save cycles, and rows without timestamps.

Suite 1763 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `29bd0876`.

Correct, and a good catch on a window I had not considered at all. I reasoned about restoration entirely in terms of *restart*, and missed that the durability of the two records diverges before that: the fast agent-state snapshot can land and the process exit before background persistence, so history holds the trigger while the display log holds nothing. After restart the block is no longer authenticatable, so the turn is stuck as a user message permanently — not degraded for one save, permanently.

The stamped row is now written up front, keyed to the prompt part's own timestamp so restoration matches the occurrence. It mirrors the social pre-save directly above it, and is best-effort for the same reason: post-processing writes the same row, so this only has to win the race against an early exit.

Also merged `main` into this branch — #169 landed the `make_user_prompt_part` choke point and its AST guard, which this branch would otherwise have violated.

5 tests: the row is written with the stamp and timestamp, what the pre-write stores satisfies the matcher end to end, no duplicate on repeat, non-fatal when the database is down, and skipped when there is no trigger.

Suite 1789 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed. I have **reverted** the prewrite rather than repairing it, in `a0b02c99`.

**Heartbeat leakage (P2):** correct, and decisive. `_persist_state` sets `skip_messages=is_heartbeat` deliberately and `_rollback_heartbeat_messages` only runs after a successful `HEARTBEAT_OK`, so a heartbeat run that errors, is cancelled, or exits early would have left the full heartbeat prompt in the user-visible log with nothing to remove it. Leaking internal text into someone's transcript is plainly worse than a cron row rendering as a user message in a rare crash window.

**Revision race (P2):** also correct — an unrevisioned read-modify-write racing the previous turn's finalizer, which can replace `messages` wholesale and drop the row regardless.

Both are fixable, but only properly by carrying the row inside the snapshot transaction. That is a persistence-layer change, not something to bolt onto this function, and this PR is not the place to attempt it — my last six revisions here have each introduced the next finding, and the two above were introduced by the revision before this one.

So: prewrite gone, and the window is now documented on `_preserve_display_triggers` along with what was tried and why it was withdrawn, so whoever picks it up does not rediscover the heartbeat trap.

**What the PR still does**, which is the common case and unaffected by any of this: provenance is stamped when the turn actually runs from an authenticated block, and restoration matches it per turn by timestamp on restart. What it no longer claims to handle is a crash between the fast state snapshot and background persistence.

I would rather ship a smaller correct change with a known limitation recorded than keep growing this one. Suite 1784 passed, ruff clean.

cyzus:
@codex review

PR #169, opened 2026-09-02
fix(security): route every user prompt through one sanitizing constructor

### LINKED ISSUE #166: Audit every entry point into the agent trust boundary
Follow-up from #162.

That PR was built assuming `ChatProcessor.process_turn` was *the* ingress. It was not — review found two more paths that reached the model without passing through it:

- `process_steer` appended steering text straight to history with `message_content=""`, so ingress sanitizing never saw it.
- `stream_acp_turn` derived the transcript from a stripped copy while passing the raw text to `_stream_prompt`, so the audit log and the executed prompt disagreed.

Both are fixed. The concern is that they were found one at a time by review rather than by enumeration, so there is no basis for believing the list is complete.

## Ask

Enumerate every path that can put text in front of the model — social, scheduled/cron, background agents, subagent delivery via `agent_inbox`, MCP, anything reaching `_stream_prompt` or `process_turn` directly — and confirm each sanitizes at ingress and derives its transcript from the same string it executes.

Worth considering a single choke point that all entry paths must call, so this cannot be reintroduced by adding a new one.

cyzus:
Closes #166.

## Why

Three entry points reached the model without sanitizing, and every one was found by review rather than by enumeration:

- `process_steer` appended steering text straight to history with `message_content=""`
- `stream_acp_turn` derived its transcript from a different string than it executed
- forking replayed stored display text

All three are fixed, but nothing suggested the list was complete — which is the actual issue. #166 asked for an enumeration plus a choke point so a new entry path can't reintroduce this.

## The enumeration

Every construction of `UserPromptPart` in `src/`:

| Site | Status before |
|---|---|
| `chat_processor.py:867` main turn | sanitized at ingress |
| `chat_processor.py:1556` steering | sanitized via `build_steering_text` |
| `context_compressor.py:717` summary request | fixed literal, no user text |
| `fork.py:46` fork from display rows | **unsanitized** |

Plus the two paths that don't construct one directly but feed the same boundary — ACP (`_stream_prompt`) and everything reaching `process_turn` (`a2a/executor`, `a2ui_routes`, `suzent_channel_routes`) — both of which are covered by their own ingress sanitizing.

`fork.py` is the one real gap this found: it rebuilds history from stored display rows, and rows from chats predating #162 can still contain raw delimiters.

## The choke point

`make_user_prompt_part()` builds every `UserPromptPart` and sanitizes on the way.

The part that matters is the guard: `test_no_module_builds_a_user_prompt_part_directly` scans `src/` and fails if any module constructs one directly. The helper on its own would just be a convention someone forgets — this is what keeps it closed. **I verified the guard fails on a planted violation** rather than assuming it would, having twice shipped tests in #162 that passed straight through the bug they were meant to catch.

## On `runtime_authored`

It's set only where this process wrapped the reminder itself, one line earlier. It says *who assembled the string*, not that the text is safe — a replayed token is still escaped, and there's a test for exactly that, since conflating the two is what caused the round-15 bearer-token bug.

## Tests

8 cases in `tests/core/test_user_prompt_choke_point.py`: the guard itself, plain and multimodal sanitizing, media passing through untouched, runtime-authored preserving a genuine reminder, runtime-authored still escaping a replayed token, and non-text content tolerated.

Suite 1756 passed, ruff clean. Ruff also caught a `NameError` I'd introduced in the steering path — the import wasn't in that scope.

### DIFF
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -726,7 +726,7 @@ class ChatProcessor:
                     )
 
         # 6. Prepare Prompt or Resume
-        from pydantic_ai.messages import ModelRequest, UserPromptPart
+        from pydantic_ai.messages import ModelRequest
 
         if message_history is None:
             message_history = []
@@ -798,6 +798,7 @@ class ChatProcessor:
         # --- System Reminder Injection (includes per-turn RAG hook when memory enabled) ---
         from suzent.core.system_reminder import (
             build_combined_reminder,
+            make_user_prompt_part,
             sanitize_untrusted_text,
         )
 
@@ -863,7 +864,7 @@ class ChatProcessor:
                 content = [full_prompt, *agent_images]
             else:
                 content = full_prompt
-            parts = [UserPromptPart(content=content)]
+            parts = [make_user_prompt_part(content, runtime_authored=True)]
             new_request = ModelRequest(parts=parts)
             message_history.append(new_request)
 
@@ -1548,11 +1549,13 @@ class ChatProcessor:
             await remove_pending_approvals(chat_id, cancelled_tool_call_ids)
 
         # 4. Append steering message
-        from pydantic_ai.messages import ModelRequest, UserPromptPart
+        from pydantic_ai.messages import ModelRequest
+
+        from suzent.core.system_reminder import make_user_prompt_part
 
         steering_text = build_steering_text(steer_message)
         message_history.append(
-            ModelRequest(parts=[UserPromptPart(content=steering_text)])
+            ModelRequest(parts=[make_user_prompt_part(steering_text)])
         )
 
         # 5. Start new agent run via process_turn with pre-built history
--- a/src/suzent/core/context_compressor.py
+++ b/src/suzent/core/context_compressor.py
@@ -710,15 +710,13 @@ class ContextCompressor:
             return messages[:1] + messages[end_index:]
 
         # Phase 6 — improved synthetic summary framing
-        from pydantic_ai.messages import UserPromptPart
+        from suzent.core.system_reminder import make_user_prompt_part
 
         summary_request = ModelRequest(
             parts=[
-                UserPromptPart(
-                    content=(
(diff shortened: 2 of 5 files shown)


cyzus:
@codex review

cyzus:
Confirmed and fixed in `79479f07`.

Correct. I exempted the whole module on the reasoning that it references `UserPromptPart` in imports and `isinstance` checks — but the pattern matches constructor calls, so neither of those trips it. Checked: `chat_processor.py` had **zero** matches, so the exemption bought nothing and blanket-excused the one file most likely to grow a new prompt path.

Now only the helper's own module is exempt, and that exemption is bounded — a second construction site there fails too. Added a test pinning why no module needs a blanket pass (`import` and `isinstance(part, UserPromptPart)` don't match; `UserPromptPart(content=x)` does), so the reasoning is checked rather than left in a comment.

Verified by planting a raw construction in `chat_processor.py` and confirming the guard now catches it at that path.

Suite 1758 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `3978d18e`.

**TextContent items (P2):** correct. `TextContent` is a string tagged with metadata and its `content` field is what reaches the LLM, but the helper treated every non-`str` item as opaque media. So the choke point had a hole in the one code path added to close holes. Clean items are returned unchanged rather than rebuilt.

**Line-based scanning (P2):** correct, and the more useful of the two. A call split across lines and an aliased import are both valid Python that the regex could not see, so the guard would stay green while production code built an unsanitized prompt — a guard blind to the thing it guards against is worse than none, because it justifies not looking. It now parses each module, resolves the names bound to the class including aliases, and walks `Call` nodes.

I verified this the way I verified the original: planted an aliased, line-split construction in `fork.py` and confirmed the guard reports it (`core/fork.py:208: constructs UPP`). The previous regex version does not catch that case, which is the point.

6 new tests: synthetic fixtures for split calls and aliases, imports and `isinstance` still ignored, plus `TextContent` sanitized and clean instances passed through by identity.

Suite 1764 passed, ruff clean.

cyzus:
@codex review

PR #171, opened 2026-09-02
fix(display): commit the trigger row inside the snapshot transaction

### LINKED ISSUE #170: Carry the trigger display row inside the state snapshot transaction
Residual from #168, documented on `_preserve_display_triggers`.

## The window

A reminder-only turn's display row is written by the background `_persist_state`. If a process takes its fast agent-state snapshot and then exits before that persist runs, the trigger is durable in history but has no stored row. After restart the block is no longer authenticatable, so nothing can vouch for the turn and it renders as a user message from then on — permanently, not for one save.

## What was tried and withdrawn

Writing the row eagerly at turn start (#168, reverted). Two problems, both found in review:

1. **It bypassed `skip_messages=is_heartbeat`.** `_persist_state` sets that deliberately and `_rollback_heartbeat_messages` only runs after a successful `HEARTBEAT_OK`, so a heartbeat run that errored, was cancelled, or exited early left the full heartbeat prompt in the user-visible log with no path that removes it. Leaking internal text is worse than the cosmetic bug being fixed.
2. **It raced the previous turn's finalizer.** An unrevisioned read-modify-write against `messages`; a stale finalizer passing its revision check afterwards replaces the list and drops the row anyway.

Anyone attempting this again should expect both.

## Suggested approach

Carry the row inside the snapshot transaction so provenance and history become durable together, rather than writing it separately and hoping the orders line up. That means touching the persistence layer (`_persist_agent_state_snapshot` / `finalize_state_if_revision_matches`), and it needs to respect the heartbeat exclusion.

## Priority

Low. Cosmetic, needs a crash in a specific window, and only affects scheduled-task rows. Filed so the analysis is not lost.

cyzus:
Closes #170 — the residual #168 deliberately left open.

## Problem

A reminder-only turn's display row was written only by the background persist. If a process took its fast agent-state snapshot and then exited before that ran, the trigger was durable in history with no stored row — and since the reminder block stops being authenticatable after a restart, nothing could vouch for the turn. It showed as a user message permanently, not for one save.

## Why the obvious fix doesn't work

#168 tried writing the row eagerly at turn start and reverted it. Two problems, both found in review there:

1. It bypassed `skip_messages=is_heartbeat`, which `_persist_state` sets deliberately. `_rollback_heartbeat_messages` only runs after a successful `HEARTBEAT_OK`, so a heartbeat run that errored, was cancelled, or exited early left the full heartbeat prompt in the user-visible log with nothing to remove it.
2. It was an unrevisioned read-modify-write racing the previous turn's finalizer, which can replace `messages` wholesale and drop the row anyway.

## Approach

Commit the row where the history it describes is already committed. `commit_snapshot_state` takes `append_display_messages` and appends them in the same transaction that writes `agent_state` and advances `state_revision`.

That closes both failure modes at once rather than one at a time:

- **No window.** Row and history become durable together or not at all.
- **No race.** Appending advances the revision, so a stale finalizer fails its check instead of clobbering `messages`. There's a test that drives exactly that sequence.

The heartbeat exclusion is now `trigger_rows_for_snapshot()` rather than a condition inline — the previous attempt got this wrong and shipped a leak, so it has a test instead of a comment.

## Tests

`tests/core/test_snapshot_carries_trigger_row.py`, 11 cases:

- rows ride the snapshot transaction, and no second unrevisioned write happens
- row and revision commit together; a stale finalizer cannot drop the row
- appending is idempotent; existing rows are kept; no rows leaves `messages` untouched
- cron yields a stamped row, **heartbeat yields nothing**, ordinary turns yield nothing, a part without a timestamp still yields a row

Also updated two existing stubs in `test_chat_processor_postprocess_persistence.py` for the new signature. Worth noting they failed *silently* first — the `TypeError` was swallowed by the broad handler around the snapshot, which is its own small hazard.

Suite 1795 passed, ruff clean.

### DIFF
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -823,6 +823,9 @@ class ChatProcessor:
             else None
         )
         pending_trigger_rows: list[dict] = []
+        # Shared with the streaming layer so the draft row this run writes can
+        # be told apart from a previous turn's answer when placing the trigger.
+        stream_run_id = str(uuid.uuid4())
         display_trigger = None
         if system_reminders and not (message_content and message_content.strip()):
             display_trigger = "\n\n---\n\n".join(
@@ -1060,6 +1063,7 @@ class ChatProcessor:
                 deferred_tool_results=deferred_tool_results,
                 permission_resolutions=permission_resolutions,
                 is_heartbeat=is_heartbeat,
+                run_id=stream_run_id,
             ):
                 try:
                     if chunk.startswith("data: "):
@@ -1172,6 +1176,7 @@ class ChatProcessor:
                         model_id=getattr(agent, "_model_id", None),
                         tool_names=getattr(agent, "_tool_names", []),
                         append_display_messages=pending_trigger_rows,
+                        draft_run_id=stream_run_id,
                     )
             except Exception as e:
                 logger.warning(
@@ -1965,6 +1970,7 @@ class ChatProcessor:
         model_id: Optional[str],
         tool_names: List[str],
         append_display_messages: Optional[List[Dict[str, Any]]] = None,
+        draft_run_id: Optional[str] = None,
     ) -> Optional[int]:
         """Persist only agent_state for fast-follow turn recovery.
 
@@ -1981,7 +1987,10 @@ class ChatProcessor:
                 messages, model_id=model_id, tool_names=tool_names
             )
             revision = db.commit_snapshot_state(
-                chat_id, agent_state, append_display_messages=append_display_messages
+                chat_id,
+                agent_state,
+                append_display_messages=append_display_messages,
+                draft_run_id=draft_run_id,
             )
             if revision is None:
                 db.update_chat(chat_id, agent_state=agent_state)
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -2689,10 +2689,18 @@ def trigger_rows_for_snapshot(
     """
     if not display_trigger or is_heartbeat:
         return []
+
+    from suzent.core.system_reminder import sanitize_untrusted_text
+
+    # Same treatment wrap_in_system_reminder gives it. Scheduled prompts and
+    # subagent results are user-influenced, and _preserve_display_triggers
+    # prefers this stored content over the placeholder rebuilt from history — so
+    # storing the raw value would put delimiters back into the visible transcript
+    # permanently, undoing the sanitizing on the path that does it.
     stamp = getattr(part, "timestamp", None)
     row: dict = {
         "role": "system_triggered",
-        "content": display_trigger,
+        "content": sanitize_untrusted_text(display_trigger),
         "trigger_origin": "runtime",
     }
     if stamp:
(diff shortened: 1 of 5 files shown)


cyzus:
@codex review

cyzus:
Both confirmed and fixed in `fb3ead6c`.

**Ordering (P2):** correct. Streaming has already flushed the turn's draft as the last stored message by the time the snapshot runs, so appending put the trigger *after* its own answer — and precisely in the recovery case this PR exists for, that is the ordering the reloaded transcript keeps. The row now goes before a trailing assistant row.

Worth flagging: my first attempt stepped over *every* trailing assistant message, which hoisted the trigger above answers belonging to earlier turns. My own test caught it, and it would have been a worse bug than the one being fixed — so it is now the final row only, with a test pinning that earlier turns keep their place.

**Derived state (P2):** correct, and I had not thought about it at all. Turning this into a durable content write means it owes what `update_chat` does — the message summary in `config` and the FTS index — otherwise a search for the trigger text permanently misses the chat and the sidebar preview stays on the previous message whenever post-processing never runs. Both are now updated in the same transaction.

4 new tests: trigger lands before the streamed answer, earlier turns keep their position, the appended row is findable via `list_chats(search=...)`, and the summary is refreshed.

Suite 1799 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `765d011e`.

Correct. A trailing assistant row is not necessarily this turn's draft — a run cancelled or failed before emitting any draft event leaves the *previous* turn's answer last, and stepping over that inserts the trigger ahead of an older answer. That corrupts precisely the chronology this placement exists to fix, which makes it worse than the bug I was fixing. I had reached for the cheapest available signal (role) twice in a row here.

The streaming layer already stamps every draft with `_streaming_run_id`, so `stream_agent_responses` now accepts a caller-supplied `run_id` and the snapshot passes the same value down. The row is stepped over only when it is a live draft carrying that id — and with no id supplied, nothing is stepped over, so any other caller keeps the conservative behaviour.

3 new tests for the cases a role check cannot distinguish: a previous turn's finalized answer, a stale draft from a *different* run (two consecutive turns losing post-processing), and no run id at all.

Suite 1802 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `f94b1d7c`.

Correct. `wrap_in_system_reminder` sanitizes the display trigger — I added that in #162 precisely because fragments are built from user-influenced material — and then here I stored the raw value on a *new* path that did not go through it. Since `_preserve_display_triggers` prefers the stored content over the placeholder rebuilt from history, delimiters in a scheduled prompt or subagent result would reappear in the visible transcript and stay there.

This is the same failure as several earlier rounds: I added a second path to the same data and did not give it the treatment the first path has. Sanitizing at the wrap point was supposed to make that impossible to forget, and it does — for anything that goes through wrapping. This does not.

2 tests: forged delimiters are neutralized before the row is stored, and ordinary trigger text is untouched.

Suite 1804 passed, ruff clean.

cyzus:
@codex review

PR #172, opened 2026-09-02
feat(skills): dedupe the catalog by revision, not by rendered lines

cyzus:
P1-4 from `docs/03-developing/system-prompt-and-reminder-audit.md`.

## Problem

`skills_reminder_hook` decided whether the model had already been told by searching history for each fully rendered line:

```python
line = f"- {skill.id}: {description} (Name: {name}; Location: {location})"
if line not in history_text:
    new_lines.append(line)
```

That fails in both directions:

- **Re-sends when nothing changed.** Reword one description and every line stops matching, so the whole catalog goes out as if new. Same for a switch between sandbox and host paths, which changes `Location:` on every line.
- **Re-sends forever.** A line differing cosmetically from what history holds never matches, so it is re-injected every single turn.

It also coupled a correctness check to formatting.

## Approach

The reminder carries a short revision marker — `[skills-catalog rev=<12 hex>]` — fingerprinting what the catalog *says* rather than how it is worded: skill ids, descriptions, and the sandbox flag that decides how locations render. Entries are sorted, so loader ordering can't cause a spurious re-injection.

## Why history, not a store

This was the real design question. A per-chat "already told" table would be the obvious move and is **worse**, because it drifts from what the model can actually still see:

- context compaction drops old turns
- reminder blocks written by an earlier process are removed from restored history (#162)

In both cases the catalog leaves the context while a store would still say "told" — so skills the agent was advertised long ago become silently unavailable, with no error and nothing to diagnose. The marker travels with the message, so it disappears exactly when the catalog does and the next turn re-advertises. Self-healing by construction rather than by bookkeeping.

There's a test for that specific property (`test_a_dropped_marker_re_advertises`).

## Verified against the #162 machinery

The marker is plain text with no delimiters, so ingress sanitizing leaves it untouched and it survives `wrap_in_system_reminder` intact — checked both, since a mangled marker would silently disable dedupe.

## Tests

14 cases in `tests/skills/test_skills_reminder_dedupe.py` — **this hook had no tests at all before**. Covers first advertisement, disabled and empty catalogs, unchanged catalogs staying quiet, adding a skill, rewording a description, loader reordering *not* counting as a change, sandbox→host switching, unrelated history not suppressing the catalog, dropped markers re-advertising, and the revision function's stability and sensitivity.

## Still open

Top-k candidate matching for large catalogs (also P1-4) is not attempted here — this change is about correctness of the dedupe, not about catalog size.

Suite 1798 passed, ruff clean.

### DIFF
--- a/src/suzent/core/system_reminder.py
+++ b/src/suzent/core/system_reminder.py
@@ -718,6 +718,37 @@ def extract_system_reminder_content(text: str) -> str:
     return "\n\n".join(parts)
 
 
+# Fragments inside one reminder block, as build_combined_reminder joins them.
+FRAGMENT_SEPARATOR = "\n\n---\n\n"
+
+_ANY_XML_EXTRACT_RE = re.compile(
+    rf"<{REMINDER_TAG}(?: nonce=\"{_NONCE_PAT}\")?>(.*?)</{REMINDER_TAG}>",
+    re.DOTALL | re.IGNORECASE,
+)
+
+
+def iter_reminder_fragments(text: str) -> list[list[str]]:
+    """Provider fragments of each reminder block in *text*, block by block.
+
+    Callers that need to find their own fragment should use this rather than
+    scanning lines: the layout has more moving parts than it looks. A block may
+    be preceded by the user's message text in the same part, several blocks may
+    be concatenated from different turns, and a reminder-only turn prefixes a
+    display-trigger envelope inside the wrapper before the body. Re-deriving all
+    of that at each call site is how a fragment ends up unfindable.
+
+    Returns one list of fragments per block, in the order the blocks appear.
+    """
+    if not text:
+        return []
+    bodies = _ANY_PUA_EXTRACT_RE.findall(text) + _ANY_XML_EXTRACT_RE.findall(text)
+    blocks = []
+    for body in bodies:
+        body = _DISPLAY_TRIGGER_RE.sub("", body)
+        blocks.append([f for f in body.split(FRAGMENT_SEPARATOR) if f.strip()])
+    return blocks
+
+
 def extract_system_reminder_display_trigger(text: str) -> str:
     """Return user-visible trigger text explicitly marked inside reminders."""
     if not text:
--- a/src/suzent/core/system_reminder.py
+++ b/src/suzent/core/system_reminder.py
@@ -741,9 +741,18 @@ def iter_reminder_fragments(text: str) -> list[list[str]]:
     """
     if not text:
         return []
-    bodies = _ANY_PUA_EXTRACT_RE.findall(text) + _ANY_XML_EXTRACT_RE.findall(text)
+    # Ordered by where each block occurs, not by which pattern found it.
+    # Concatenating the two result sets put every PUA block before every XML one,
+    # so a history spanning a format switch reported the older block as newest.
+    found = [
+        (match.start(), match.group(1))
+        for pattern in (_ANY_PUA_EXTRACT_RE, _ANY_XML_EXTRACT_RE)
+        for match in pattern.finditer(text)
+    ]
+    found.sort(key=lambda item: item[0])
+
     blocks = []
-    for body in bodies:
+    for _, body in found:
         body = _DISPLAY_TRIGGER_RE.sub("", body)
         blocks.append([f for f in body.split(FRAGMENT_SEPARATOR) if f.strip()])
     return blocks
@@ -873,8 +882,15 @@ async def build_combined_reminder(
         logger.debug(f"[system-reminder] chat={chat_id} — no content, skipping")
         return None
 
+    # The separator is in-band, so a fragment carrying it verbatim would split
+    # into two and let its own content pose as a separate provider's. Goal and
+    # task text is unrestricted, so collapse the blank lines that make it a
+    # boundary: the rule stays visible and no characters are lost.
     result = wrap_in_system_reminder(
-        "\n\n---\n\n".join(parts), display_trigger=display_trigger
+        FRAGMENT_SEPARATOR.join(
+            part.replace(FRAGMENT_SEPARATOR, "\n---\n") for part in parts
+        ),
+        display_trigger=display_trigger,
     )
     # Metadata only. The wrapped text carries RUNTIME_NONCE, and file logging
     # records DEBUG unconditionally — writing it to disk would hand the token to
--- a/src/suzent/core/system_reminder.py
+++ b/src/suzent/core/system_reminder.py
@@ -725,9 +725,15 @@ _ANY_XML_EXTRACT_RE = re.compile(
     rf"<{REMINDER_TAG}(?: nonce=\"{_NONCE_PAT}\")?>(.*?)</{REMINDER_TAG}>",
     re.DOTALL | re.IGNORECASE,
 )
+_OWN_XML_EXTRACT_RE = re.compile(
+    rf"<{REMINDER_TAG} nonce=\"{RUNTIME_NONCE}\">(.*?)</{REMINDER_TAG}>",
+    re.DOTALL | re.IGNORECASE,
+)
 
 
-def iter_reminder_fragments(text: str) -> list[list[str]]:
+def iter_reminder_fragments(
+    text: str, *, authenticated_only: bool = False
+) -> list[list[str]]:
     """Provider fragments of each reminder block in *text*, block by block.
 
     Callers that need to find their own fragment should use this rather than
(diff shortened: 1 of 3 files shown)


cyzus:
@codex review

cyzus:
Both confirmed and fixed in `7eba1bbc`.

**Stale marker match (P1):** correct, and the more interesting of the two. A→B→A stayed silent because history still held A's marker, while what the model had most recently been told was B. Only the last marker counts now. There is a test walking that exact sequence.

**Locations missing from the fingerprint (P2):** correct. I chose which inputs to cover — ids, descriptions, sandbox flag — and a path can change on its own, giving the same revision for a different `Location`.

Rather than add locations to the input list, the revision now hashes **the rendered lines themselves**. Picking inputs means keeping that list in sync with the rendering forever, and I had already got it wrong once; fingerprinting the output cannot fall behind. That is the same lesson as `sanitize_tool_payload` in #162 — check what is actually emitted, not a model of what goes into it.

Also simplified the hook: lines are built once and used for both the fingerprint and the message, so there is no second place for the two to diverge.

8 new tests: order independence, revision sensitivity to description/added-skill/location, latest-marker selection, no-marker, A→B→A re-advertising, and a moved skill re-advertising with its new path.

Suite 1802 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `fa6b1503`.

**Marker scope (P2):** correct. A bare marker pattern matches marker-shaped text anywhere in the prompt — a repository reminder, a goal, something a user pasted — and because the newest match wins, that text could decide whether the catalog is advertised. Worse in one direction than the other: matching a *current* revision suppresses the catalog entirely.

The header is now single-sourced as `CATALOG_HEADER` and used both to emit and to match, so recognition is scoped to something this hook actually wrote, and the emitted text cannot drift from the pattern that recognises it. Two tests: a bare marker is not recognised, and a marker pasted into a user message does not suppress the catalog.

**Missing annotations (P1):** correct, `AGENTS.md:17` says to use type hints everywhere and I did not. Annotated the helpers and the cases. Worth noting the sibling files in `tests/skills/` annotate parameters but no return types — I went with full annotations to match the stated rule rather than the older files, so if the house preference is actually the looser one, say and I will match it instead.

Suite 1804 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `d4ac5670`.

Correct, and pointed: the header anchoring I added last round to stop one variant of marker confusion **reintroduced the runaway repetition this PR exists to fix**. Descriptions and locations render *after* the real marker, so metadata reproducing the header plus a marker becomes the 'latest advertised revision', never matches the true one, and the catalog re-injects every single turn.

Fixed at the source: marker and header shapes in rendered fields carry a word joiner, so the text reads identically and no longer matches the pattern. Content is preserved rather than dropped — this is a rendering concern, not a trust boundary.

**One residual I want stated rather than glossed.** That closes the reachable case, since skill metadata is the only content this hook renders. A foreign reminder fragment reproducing `CATALOG_HEADER` verbatim — a repository AGENTS.md documenting it, say — would still be read as the latest advertisement. I have left that rather than fixed it, because the failure direction is asymmetric: it causes **re-advertising**, not silence, so the model's catalog stays correct and the cost is a repeated reminder. Choosing 'any marker matches' instead would trade that for the A→B→A silence you flagged earlier, which leaves the model with a stale catalog. I would rather be noisy than wrong, but if you read the trade-off differently say so, because it is a judgement call and not a fact.

It is noted in `latest_advertised_revision`'s docstring so the next reader sees the reasoning rather than assuming it is airtight.

3 tests: a hostile description and a hostile location both settle after one advertisement instead of looping, and neutralizing keeps the text readable.

Suite 1807 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `addd74c7`.

Correct, and my justification for the previous fix was wrong. I called mangling the text 'a rendering concern, not a security boundary' — but a rendered identifier is *functional data*: `SkillTool` needs an exact id, and `SkillMetadata.name` is unrestricted, so a skill whose id or path contained `skills-catalog rev=` became unloadable. I traded a token-waste bug for a broken skill.

No catalog data is altered now. The marker is recognised **positionally** — a line of its own directly beneath a line that is exactly the header — and entries are folded onto single lines, which is all that is needed to stop catalog content contributing a line of either shape. Only newlines are touched; ids, names and paths never contain them and reach the model unchanged.

That also retires the residual I flagged last round, and better than I expected: a foreign fragment would now need to place the exact header on its own line with a bare marker line directly beneath it, rather than merely contain the header string.

4 tests: identifiers with the marker substring pass through verbatim, a multi-line description cannot forge a block, a marker sharing a line with other text is ignored, and a marker without the header above it is ignored.

Suite 1810 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `e304156f`. This is the residual I told you was 'retired' last round — that claim was wrong, and you found the path.

`goal.objective` is unrestricted multi-line text, and `plan_reminder_hook` is registered after this hook, so its fragment lands after the real marker. A header-plus-marker pair anywhere in that text was read as the current revision, never matched, and re-injected the catalog every turn.

The header must now be the **first runtime-authored line of a reminder fragment**, with the marker directly beneath. Fragment position is the thing a lookalike cannot reach: the plan fragment begins with its own `[ACTIVE GOAL]` line, so text inside it is never at a boundary.

Two things I got wrong on the way, both worth recording:

**My tests were the reason this kept slipping through.** They fed a bare fragment as history. Real history is user text, then the wrapper, then our fragment, then other providers' — so extraction has to start after the wrapper opening, or the header is never first and the marker is never found *at all*. My first version of this fix would have disabled dedupe completely in production while every test passed. The samples now build history the way a turn actually stores it, which is why the goal case is now expressible as a test at all.

**One of my string replacements silently did nothing.** I patched that test without asserting the match landed, ruff had re-wrapped the line, and the old body survived. The failure looked like a logic bug in the fix; it was an unapplied edit. Every replacement in this session's later edits asserts `count == 1` first, and this one now does too.

3 new tests: the marker is found in realistic history alongside user text and a plan fragment, a goal reproducing the header does not hijack the revision, and the latest of two turns wins.

Suite 1812 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `ac04dcea`. Both were in code I wrote last round, and together they say the approach was wrong rather than incomplete.

**Concatenated blocks (P1):** reminder blocks from separate turns are joined with a newline, and my parser reset its line buffer at every wrapper opening — so it kept only the *final* block. A catalog turn followed by a plan-only turn lost the marker entirely and re-injected from then on.

**Display-trigger envelope (P2):** reminder-only turns prefix it inside the wrapper, so the header was not the first line and scheduled turns never recognised their own catalog. Exactly the population that runs most often.

Both come from the same root cause: I re-derived the reminder layout inside the skills hook from line heuristics, and the layout has more moving parts than it looks — user text before the block, several blocks concatenated, a display-trigger envelope inside the wrapper, provider fragments within the body. I got a different piece of it wrong on each of the last three rounds.

So the hook no longer guesses. `system_reminder` exposes `iter_reminder_fragments()`, which returns each block's provider fragments with the envelope stripped, and the hook consumes that. Format knowledge now lives with the module that owns the format, which is where it should have been from the start.

Worth noting what let this run for three rounds: my tests fed **bare fragments** as history. Every variant passed while production would have mis-parsed. The samples build history as a turn stores it, and there are now explicit cases for concatenated turns and a reminder-only turn — the two shapes I had never exercised.

3 new tests plus two corrected. Suite 1815 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `2e08f5f2`.

**Block ordering (P2):** correct. I collected blocks by concatenating the two pattern results, so every PUA block sorted before every XML one regardless of position — a history spanning a switch to or from the XML debugging fallback reported the older block as newest. Merged by source offset now.

**Separator injection (P2):** correct, and the more structural of the two. The separator is in-band, so a fragment carrying it verbatim splits into two and its own content poses as another provider's — and goal and task text is unrestricted. Fragments now have the separator's blank lines collapsed before joining: the boundary is destroyed, the rule stays visible, no characters are lost. Deliberately not the word-joiner approach from earlier in this PR, since that corrupted data the model has to copy.

**I verified both tests fail against the unfixed code** rather than trusting that they exercise the change — for the separator case that meant reverting the escaping and watching the forged revision win. Worth doing here specifically because two earlier rounds on this PR had tests that passed while production was broken.

One process slip: I used `git checkout --` to undo those experiments and discarded my own uncommitted fixes with them, then briefly reported a clean suite that was actually testing unfixed code. Caught it on the next run and re-applied. Committing before experimenting would have avoided it.

2 tests. Suite 1817 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `b19bce1d`.

Correct, and it is an ordering relationship I had not considered at all: this hook runs **before** the history processor strips unauthenticated blocks. So on the first turn after a restart the previous process's catalog is still sitting in history, my extractors accepted it, the advertisement was suppressed — and then the processor removed that very block, leaving the model with neither the old catalog nor a replacement for that turn. The one moment the catalog is most needed.

`iter_reminder_fragments` now takes `authenticated_only`, matching only blocks signed with the current `RUNTIME_NONCE`, and the hook asks for it. Re-advertising once per restart is the intended behaviour, and it now matches how the rest of the system treats unauthenticated history rather than quietly diverging from it.

Note this makes the earlier reasoning in the PR description true rather than aspirational: I argued history was the right source because the marker 'disappears exactly when the catalog does'. That was only correct if unauthenticated blocks are excluded, which they were not until now.

3 tests: a previous process's identical catalog does not suppress, an untokenized legacy block does not suppress, and block ordering is still exercised at the extractor level where fake nonces are appropriate.

**Verification note:** my first attempt to prove the tests fail without the fix stashed the tests along with the source, so it ran the old suite and proved nothing — I reported it as meaningless and redid it stashing only `src/`. Both new tests do fail against the unfixed code.

Suite 1820 passed, ruff clean.

cyzus:
@codex review

PR #173, opened 2026-09-02
fix(security): annotate the sanitizer's RunContext so it is actually invoked

cyzus:
Fixes a runtime error introduced in #162.

```
make_tool_output_sanitizer_history_processor.<locals>._processor()
missing 1 required positional argument: 'messages'
```

## Cause

pydantic-ai decides whether to pass a `RunContext` by inspecting the **first parameter's type annotation** (`_utils.takes_run_context`, which checks `first_param_type is RunContext or get_origin(...) is RunContext`).

The sanitizer annotated it `ctx: Any`, so pydantic-ai classified it as the messages-only form and called `_processor(messages)` — binding the message list to `ctx` and leaving `messages` unfilled. Every model request raised `TypeError`, and **no tool-output sanitizing ran at all**.

`make_compaction_history_processor` annotates `RunContext[AgentDeps]` and was unaffected, which is why only one of the two broke.

## Why the tests didn't catch it

They called `processor(None, [...])` directly — bypassing precisely the dispatch that was broken. 110 tests on this module, all green, none of them exercising how the thing is actually invoked. Same failure mode as three earlier rounds on #172, where tests fed bare fragments while production passed something else.

So the regression tests go through the real entry point:

- `test_processor_runs_through_pydantic_ai_dispatch` calls `_run_history_processor`, the function the agent uses, so the ctx/no-ctx decision is exercised rather than assumed.
- `test_processor_declares_that_it_takes_a_run_context` pins the annotation directly, so loosening it fails loudly rather than silently reverting to the messages-only path.
- `test_every_registered_history_processor_takes_a_run_context` checks both registered processors, since the two disagreed and only one was wrong.

All three fail against the unfixed annotation — verified by reverting it and re-running.

## Impact

Tool-output sanitizing was inert wherever this shipped. Ingress sanitizing (chat, ACP, steering, fork), stored-history handling, and the wrap-point cleaning were unaffected — those do not go through a history processor. Model-output sanitizing shares this processor and was equally inert.

Suite 1820 passed, ruff clean.

### DIFF
--- a/src/suzent/core/system_reminder.py
+++ b/src/suzent/core/system_reminder.py
@@ -22,6 +22,8 @@ import re
 import secrets
 from typing import Any, Callable, Awaitable, Optional, List
 
+from pydantic_ai.tools import RunContext
+
 from suzent.logger import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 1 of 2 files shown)


cyzus:
@codex review

PR #174, opened 2026-09-02
refactor(prompts): give each rule exactly one owning layer

cyzus:
P1-5 from the prompt/reminder audit.

## Problem

Four rules were stated in two layers each. The layers are assembled independently, so duplication is easy to introduce and invisible afterwards — and it costs twice: tokens on every turn, and two sites to update when the rule changes.

| Rule | Was in | Now owned by |
|---|---|---|
| Shell is not for file read/search/edit | `STATIC_INSTRUCTIONS` **and** `RunCommandTool.session_guidance` | the tool's guidance |
| Verification threshold (`3+ file edits`) | `STATIC_INSTRUCTIONS` **and** `AgentTool.session_guidance` | the Verification Contract |
| When to reach for a skill | `SkillTool.session_guidance` **and** the catalog reminder header | the tool's guidance |
| Diagnose before retrying | Behavioral Guidelines bullet **and** Failure Handling SOP | the SOP |

## Ownership rule applied

- global safety and authorization → `STATIC_INSTRUCTIONS`
- how to call a tool → its schema
- when to reach for a tool → that tool's `session_guidance`
- what is available right now → reminders

Two of these are more than tidying:

**Shell-vs-files** was unconditional in the static prompt, so every session paid for it including ones with no shell tool equipped. As tool guidance it appears only when the tool does.

**Skill usage** is the one that was actually broken. Since #172 the catalog reminder only fires when the skill set *changes*, so `"Use it IMMEDIATELY when the user's task matches a skill"` living in that header meant the standing instruction disappeared after the first advertisement. The header now names the resource (`Skills available to load with SkillTool:`) and `SkillTool.session_guidance` — present every turn — carries the policy.

Note the header change invalidates catalog markers in existing histories, so each live chat re-advertises once after deploy. That is the intended self-healing behaviour, not a regression.

## Tests

`tests/prompts/test_rule_ownership.py` pins each rule to its layer and caps `STATIC_INSTRUCTIONS` at 3000 chars (currently 2965) so trimmed text cannot creep back in.

**All five guards fail against the previous code** — verified by stashing `src/` and re-running.

Suite 1848 passed, ruff clean.


cyzus:
@codex review

cyzus:
Heads up: I pushed the first commit with a failing test — `test_build_session_guidance_section_tool_aware_rules` still asserted the old `"use SkillTool early"` wording. Fixed in `98784c17`.

My mistake was mechanical: I ran the suite and the commit in one shell chain separated by `;`, so the test failure did not stop the push. CI would have caught it, but I reported the suite as green in the PR description when it was not. Suite is 1848 passing now.

cyzus:
@codex review

PR #175, opened 2026-09-02
refactor(memory): move notebook procedure into the notebook skill

cyzus:
P1-6 from the prompt/reminder audit.

## Problem

The core-memory section is injected on **every turn** a chat has memory enabled. Roughly a third of it was notebook procedure:

- vault structure and "your notebook IS the wiki"
- `index.md` / `log.md` navigation
- the pre-creation GlobTool duplicate check
- the query workflow
- what counts as a durable output
- pointers to the `ingest.md` / `lint.md` runbooks

None of it applies unless someone is working in the vault, and all of it competed for attention on ordinary coding and chat turns.

## What moved where

The `notebook` skill already owned nearly all of it — schema authority, `index.md`/`log.md` rules, ingest and lint procedures, filing policy, and the Glob check (`ingest.md:53`). So this is mostly deleting a second copy.

Only two things were unique to the prompt, and those moved into `SKILL.md` rather than being dropped: the query workflow (read index → synthesize with citations → file only when authorized) and the definition of a durable output.

**Core memory keeps state and the rules that protect it:** the blocks, where the files live, and the warning that consolidation rewrites everything above `MEMORY.md`'s generated marker — losing that one would cost the user real text.

## Behaviour change worth noting

In host mode with no notebook mount, the section previously said "Load the `notebook` skill … If notebook is not configured, skip notebook operations." It now just says no notebook is configured. Naming a skill for a resource the session does not have is noise, and the skill itself already handles the unconfigured case. `test_memory_context_host_mode_avoids_virtual_paths` was asserting the old string; updated, and a new case covers the mounted-host path so pointer coverage is not lost.

## Result

**3264 → 2261 characters**, every turn.

`tests/memory/test_core_memory_scope.py` asserts the blocks survive, procedure is gone, the pointer remains, the unconfigured case is clean, the `MEMORY.md` safety rule survives, and the section stays under 2600 chars. Five of them fail against the previous version — verified by stashing `src/`.

Suite 1855 passed, ruff clean.

### DIFF
--- a/skills/notebook/SKILL.md
+++ b/skills/notebook/SKILL.md
@@ -55,6 +55,9 @@ writes.
 
 When filing is authorized:
 
-1. Follow `schema.md` for the destination and page structure.
-2. Update `index.md` only when required by the schema.
-3. Append a `query-filed` entry to `log.md` when the log convention requires it.
+1. Check what already exists first — read `index.md`, and Glob the vault for the
+   topic. A page written without looking is a duplicate or an orphan. Update or
+   link the existing page instead of creating a second one.
+2. Follow `schema.md` for the destination and page structure.
+3. Update `index.md` only when required by the schema.
+4. Append a `query-filed` entry to `log.md` when the log convention requires it.
--- a/skills/notebook/SKILL.md
+++ b/skills/notebook/SKILL.md
@@ -36,6 +36,17 @@ Use the schema's link style. Prefer full vault-relative paths when links could b
 Read the selected procedure completely before acting. Treat referenced binary assets as
 immutable unless the user explicitly asks to replace them.
 
+## Answering from the notebook
+
+1. Read `index.md` first to identify candidate pages.
+2. Read those pages and synthesize with citations back to them.
+3. File the result only under the rules below — answering a question is not
+   itself a reason to write a page.
+
+Durable outputs are comparisons, analyses, syntheses, and decision breakdowns:
+material someone would want to find again. A direct answer to a direct question
+is not one.
+
 ## Filing conversation results
 
 File a result only when the user explicitly requests it or the vault/session has an
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -500,13 +500,24 @@ class ChatProcessor:
         # the stable config, enabling a skill updates persistence but reuses a
         # cached agent that has no SkillTool — and the prompt would then point
         # at a tool the run does not actually have.
+        #
+        # Read through the same per-chat manager build_agent_deps uses, not the
+        # global singleton: toggle_skill persists through a per-chat manager and
+        # never reloads the singleton, so the singleton's answer can be stale
+        # for the life of the process.
         try:
-            from suzent.skills.manager import SkillManager
+            from suzent.skills.manager import get_skill_manager_for_chat
 
             config["_skills_enabled"] = bool(
-                SkillManager.get_instance().has_enabled_skills()
+                get_skill_manager_for_chat(
+                    chat_id,
+                    config.get("cwd"),
+                    custom_volumes=config.get("sandbox_volumes"),
+                    sandbox_enabled=config.get("sandbox_enabled", True),
+                ).has_enabled_skills()
             )
-        except Exception:
+        except Exception as e:
+            logger.debug(f"Could not read skill enablement for {chat_id}: {e}")
             config["_skills_enabled"] = False
 
         config["_has_discovered_skills"] = bool(
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -496,6 +496,19 @@ class ChatProcessor:
             if repository_context.repository_root is not None
             else None
         )
+        # SkillTool equipment is decided once, in create_agent. Without this in
+        # the stable config, enabling a skill updates persistence but reuses a
+        # cached agent that has no SkillTool — and the prompt would then point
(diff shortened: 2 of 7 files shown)


cyzus:
@codex review

cyzus:
Both confirmed and fixed in `984a4d5a`.

**Skill not loadable (P1):** correct, and I checked the premise before acting — `_load_enabled_state()` starts from an empty set, so on a fresh install no official skill is enabled and `create_agent()` never equips `SkillTool`. My change left the prompt telling the model to load a tool it does not have, having just deleted the procedure it used to rely on. That is worse than either the bloat or the removal, because a dead pointer reads as a route out of the problem.

The hint is now conditional on the skill actually being enabled. When it is not, it points at the vault's own `schema.md` — which `SKILL.md` already calls the sole authority, so the fallback reaches the same source without a tool in between — plus the two rules that prevent real damage: check before creating a page, and file only when asked. Two lines, not thirty. Unknown state resolves to unavailable, because pointing at the vault always works and pointing at an absent tool does not.

**Filing without checking (P2):** correct. The Glob check lives in `ingest.md`, which covers source consolidation only, so "file this in my notebook" could reach the filing workflow without anyone looking for an existing page. Added as step 1 of that workflow in `SKILL.md`.

6 new tests: the fallback names the vault and not the skill, stays short and free of runbook detail, and unknown skill state resolves to unavailable.

Suite 1861 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `d820b7b4` — both halves, since either alone would still be wrong.

**Wrong manager:** correct. `build_agent_deps` creates a per-chat manager via `get_skill_manager_for_chat`, whose discovered skills are enabled by default, while I consulted the global singleton. So a repository-provided `notebook` skill could be loadable through `ctx.deps.skill_manager` while the prompt told the model it had no skill — the precise inversion of the bug I was fixing, since the previous round's problem was pointing at a tool that was absent. Now reads the run's own manager.

**Stale cache:** correct, and I would have missed it. The value changes the rendered text, so it belongs in the cache identity; keyed only on memory-file revisions, a toggle would leave existing chats on the old wording until an unrelated file happened to change. Added to the key.

4 tests replacing the two that probed the singleton: availability reads the run's manager, a missing manager and a broken one both resolve to unavailable, and toggling produces different cached sections for the same chat.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `6f0013e6`.

Correct. I built the fallback around what could *damage* the vault — schema authority, check before creating, file only when asked — and left out how to *read* from it. On a fresh install that fallback is the only notebook guidance the model receives, so an explicit notebook question lost its retrieval path rather than merely degrading. Two different things go wrong in a vault, and I guarded one.

Added in one line: start at `index.md` and cite the pages used. The fallback is now 4 lines against the 30 that were there before this PR, and still points at `schema.md` for anything structural.

Test asserts both halves — answering and writing — so a future trim cannot quietly drop one again.

Suite 1865 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `e0e943dc`.

Correct. I compressed "only when the user explicitly requests it **or** the vault/session has an enabled capture policy" down to "only when asked to" and dropped a clause that authorises work. Because this text is the only notebook guidance when the skill is unavailable, that silently disabled automatic captures in precisely the configuration with nothing else to fall back on — and it contradicted `SKILL.md`, which still carries the exception.

Restored. Test asserts the exception is present, since this is now the third round where trimming the fallback removed something load-bearing: first the answering path, now the capture policy. Each time the summary read fine in isolation and lost a capability.

Suite 1865 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `e6e28d84`.

Correct. `SkillTool` equipment is decided once in `create_agent`, while enablement is read per request — so after a toggle the per-request manager says "available" and the cached agent still has no `SkillTool`. That is the same dead pointer this PR exists to remove, reached through a door I had not checked.

I fixed the cause rather than the message. Recording enablement in the stable agent config makes a toggle rebuild the agent, so the tool really is equipped and the prompt's claim becomes true. Basing the hint on "is a tool present" instead would have made the wording accurate while leaving the actual problem — a session that cannot use a skill the user just enabled — untouched.

2 tests: skill state is written into the agent config, and is not in `_TRANSIENT_KEYS` so it genuinely participates in the cache key.

Suite 1867 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `3447b409`.

Correct, and it is the same mistake as two rounds ago in this PR: I reached for `SkillManager.get_instance()` again. `toggle_skill` mutates and persists through a manager from `get_skill_manager_for_chat` and never reloads the singleton, so the singleton's answer can be stale for the life of the process — meaning my new cache discriminator read a value that does not change on toggle, and `SkillTool` stayed unequipped until restart anyway.

Earlier in this PR you pointed out that `_notebook_skill_available` had the same flaw and I moved it to `ctx.deps.skill_manager`. I then wrote a *new* singleton read into the fix for the follow-up finding. Reading it through `get_skill_manager_for_chat` now, the same accessor `build_agent_deps` uses.

Test asserts both halves — that the per-chat accessor is used, and that `SkillManager.get_instance()` does not appear — because the positive assertion alone would have passed while the stale call sat next to it.

Suite 1867 passed, ruff clean.

cyzus:
@codex review

PR #177, opened 2026-09-02
fix(plan): charge goal budget on completed turns, not on prompt assembly

cyzus:
P1-3 from the prompt/reminder audit.

## Problem

`plan_reminder_hook` incremented `turns_elapsed` while *building* the reminder:

```python
parts.append("Evaluate: if the goal is achieved call manage_goal(action='clear')...")
db.update_goal(goal.id, turns_elapsed=goal.turns_elapsed + 1)   # inside a read
```

Reminder providers run on every path that assembles a prompt. So the budget was charged by heartbeats, approval resumes, tool continuations and retries — none of which is a turn the user took. `turns_elapsed` measured how often a prompt got built, and goals hit "turn budget exhausted" early for reasons nobody could see.

## What "a turn" means now

A real user message. `process_turn` already computes exactly that when deciding whether to run per-turn hooks — non-empty content, not a heartbeat, not an approval resume — so `is_user_turn=bool(_turn_message)` reuses it rather than introducing a second definition that can drift from the first.

The count advances once, after the turn has run and persisted. A failed stream is not charged. `advance_goal_turn` is best-effort: bookkeeping must not fail a turn that already succeeded.

## Migration: none, deliberately

Existing goals keep their inflated counts. There is no way to tell retroactively which increments were real turns, so resetting would hand back budget that may genuinely have been spent. Counting is accurate from here, and the visible effect is that active goals stop consuming budget when nothing is happening.

If you would rather reset counts on deploy, that is a one-line migration — but it needs to be your call, since it changes when live goals pause.

## Tests

10 cases in `tests/core/test_goal_turn_counting.py`: building the reminder spends nothing, five reminders in a row still spend nothing, the reminder still *reports* the budget, a completed turn charges exactly one, paused goals and missing goals/projects are no-ops, a database failure cannot break the turn, and the chargeable-turn definition is the one `process_turn` already uses.

Suite 1853 passed, ruff clean.

### DIFF
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -837,13 +837,13 @@ class ChatProcessor:
             _is_user_turn if counts_toward_goal is None else counts_toward_goal
         )
 
-        from suzent.tools.plan_hooks import active_goal_id, advance_goal_turn
+        from suzent.tools.plan_hooks import active_goal_identity, advance_goal_turn
 
         # Pinned at turn start: a goal the agent creates mid-turn did not cost
         # this turn anything, and charging it can pause a max_turns=1 goal
         # before it runs a single step.
         _goal_at_start = (
-            active_goal_id(chat_id)
+            active_goal_identity(chat_id)
             if _counts_toward_goal and not is_heartbeat
             else None
         )
@@ -1265,7 +1265,7 @@ class ChatProcessor:
             # block: the judge could read the previous count, start one more
             # autonomous run, and take the goal past max_turns.
             if _counts_toward_goal and not stream_failed and _goal_at_start:
-                advance_goal_turn(chat_id, only_goal_id=_goal_at_start)
+                advance_goal_turn(chat_id, only_goal=_goal_at_start)
 
             # 10. Goal-mode continuation. After a normal turn, let the judge decide
             # whether to auto-continue toward a standing goal. Cheap no-op (one
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -444,6 +444,7 @@ class ChatProcessor:
         _message_history_override: list = None,
         system_reminders: list[str] = None,
         incoming_citation_sources: list[dict] = None,
+        counts_toward_goal: Optional[bool] = None,
     ) -> AsyncGenerator[str, None]:
         """
         Process a user message turn:
@@ -1204,7 +1205,11 @@ class ChatProcessor:
                     agent=agent,
                     postprocess_job_id=postprocess_job_id,
                     file_snapshot=file_snapshot_json,
-                    is_user_turn=bool(_turn_message),
+                    counts_toward_goal=(
+                        bool(_turn_message)
+                        if counts_toward_goal is None
+                        else counts_toward_goal
+                    ),
                 )
 
             task_id = f"post_process_{chat_id}_{postprocess_job_id}"
@@ -1277,7 +1282,7 @@ class ChatProcessor:
         agent: Any,
         postprocess_job_id: str,
         file_snapshot: list[dict],
-        is_user_turn: bool = False,
+        counts_toward_goal: bool = False,
     ) -> None:
         """Background post-processing for a completed turn.
 
@@ -1436,7 +1441,7 @@ class ChatProcessor:
                 # hook, which every prompt-assembling path calls — heartbeats,
                 # approval resumes, tool continuations and retries all spent
                 # budget on turns the user never took.
-                if is_user_turn and not stream_failed:
+                if counts_toward_goal and not stream_failed:
                     from suzent.tools.plan_hooks import advance_goal_turn
 
                     advance_goal_turn(chat_id)
@@ -1529,6 +1534,10 @@ class ChatProcessor:
             chat_id=chat_id,
             user_id=user_id,
             message_content=replay_message,
+            # The original turn already charged the goal, and the retry
+            # checkpoint restores chat state, messages and files but not the
+            # counter — so replaying would bill the same work twice.
+            counts_toward_goal=False,
             files=replay_files if replay_files else None,
             config_override=merged_config,
             is_social=is_social,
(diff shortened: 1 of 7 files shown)


cyzus:
@codex review

cyzus:
Both confirmed and fixed in `de48b70b`. The first is the more serious, and it means my fix traded an inflated budget for an **unenforceable** one.

**Autonomous goal steps (P1):** correct. `run_goal_step` calls `process_background_turn` with an empty `message_content` and only a system reminder, so `_turn_message` is None and nothing was charged — and since the hook no longer charges either, `max_turns` could never stop a goal whose judge keeps requesting more work. Over-counting wastes budget; not counting removes the stop condition entirely.

**Replayed retries (P2):** correct. The replay looks like a fresh user message, the original turn already charged, and `apply_retry_checkpoint` restores chat state, messages and files but not the counter.

The underlying mistake was picking "a real user turn" as the rule. A goal budget measures *work on the goal*, which includes autonomous continuation and excludes replays — those are different sets, and I conflated them because `_turn_message` was conveniently to hand. `process_turn` now takes `counts_toward_goal`, defaulting to None for "infer", with the two callers that know better answering explicitly. Renamed from `is_user_turn` so the parameter says what it decides rather than naming one case that happens to decide it.

4 tests replacing the single inference check: default inference, autonomous step chargeable, retry not chargeable, and the explicit override reaching `process_turn`.

Suite 1856 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `6d4b4363`.

**Broken forwarding (P1):** correct, and worse than the bug this PR is about. `run_goal_step` reaches `process_turn` via `process_background_turn` → `process_turn_text`, and I added the keyword to the call without adding it to `process_turn_text`. `TypeError` before any model work, caught and merely logged by `run_goal_step` — so every autonomous goal step would stop silently. I checked the chain by signature this time and added a test that walks all three, because the failure mode here is invisible.

**Retry after a failed original (P2):** correct, and it exposes that my round-1 fix rested on an assumption I never checked. Forcing retries uncharged assumed the first attempt had paid — but `save_retry_checkpoint` runs at turn *start*, before the outcome exists, so nothing records whether it did.

Rather than add a field that cannot be populated at the only moment it is written, retries are now charged like any other turn. The argument, since this is a judgement call and reverses what I told you last round: **for a budget whose purpose is stopping runaway work, the two errors are not symmetric.** Over-charging pauses a goal early, and `/goal resume` recovers it. Under-charging removes the stop condition, and nothing recovers that. Your round-1 point about double-billing stands — I now think paying twice for two real attempts is the lesser cost.

If you would rather have exactness here, the way to get it is recording the charge on the goal at increment time and consulting it at retry, which is a larger change than this PR should carry.

Suite 1857 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `14e6aa8a`.

**Race with continuation (P2):** correct. The increment ran in the background post-process while `process_turn` scheduled `maybe_continue_goal` inline, so the judge could read a stale count and start one more autonomous run past `max_turns`. Rather than add a wait, the increment moved inline ahead of that block — the budget the judge reads now already includes this turn, and there is one charging site instead of two that can disagree. A test asserts the ordering by position in the source.

**Attachment-only turns (P2):** correct, and it invalidates the reasoning I gave for the original design. I said reusing `_turn_message` avoided "a second definition that can drift" — but they were never the same definition. `_turn_message` answers *should retrieval hooks run*, which needs query text; chargeability answers *did the user ask for work*, which a file alone does. Sharing one value made attachment-only prompts run the agent, trigger continuation, and spend nothing. They are separate predicates now, and the test says why.

4 tests replacing 1. Suite 1859 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `148d0d38`.

**Steering (P2):** correct. `process_steer` puts its text straight into `_message_history_override` and calls `process_turn` with empty content, so my predicate said no — while the run does real work and can trigger continuation. It declares itself chargeable now. That is the third entry point where the predicate needed the caller's help rather than inference; the pattern is that anything not arriving as plain `message_content` has to say what it is.

**Goal created during the turn (P2):** correct, and the failure is worse than a miscount — at `max_turns=1` the setup turn exhausts the budget and `maybe_continue_goal` pauses the goal before it runs a single autonomous step, so asking the agent to set a goal produced a goal that never ran. The charge is pinned to the goal id read at turn start, and a goal that was not running then is not charged.

4 tests: a mid-turn goal is not charged, the pre-existing one is, `active_goal_id` tracks status, and steering declares itself.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `29912f1d`.

Correct. `_set_goal` replaces an active goal by updating the row in place and resetting `turns_elapsed`, so the id survives — my pin accepted the replacement and charged it for work done under the objective it replaced. At `max_turns=1` that exhausted the new goal before its first autonomous step, which is the same user-visible failure as the created-goal case I fixed last round, reached by a different route I did not check.

Pinned on `(id, objective)` now. Worth being explicit about the limit: replacing a goal with a byte-identical objective is indistinguishable from not replacing it, and will still charge. That seems right — nothing meaningful changed — but it is an assumption rather than a guarantee, and a monotonic version column on the goal row would make it one.

Test covers replacement with a different objective; the created-goal and normal cases still pass.

Suite 1864 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `083de91a`, this time with the identity rather than another proxy for it.

You are right that adding `objective` could not close this: re-setting a goal to change only `max_turns` leaves the text byte-identical while resetting `turns_elapsed`. That is the fourth round on this question, and my first three answers — row id, then (id, objective) — were both attempts to infer an identity the data did not actually carry.

So the data carries it now. `GoalModel` gains a `generation` counter, bumped by `_set_goal` when it replaces a goal in place and at no other time, with a migration adding the column to existing rows (defaulting to 0, so a goal that has never been replaced compares equal as before). The pin is `(id, generation)`.

Verified the column round-trips through a real `ChatDatabase`, since a migration that silently no-ops would make every comparison trivially equal and the tests would still pass.

2 tests: replacement detected, and replacement with identical objective text detected — the case that proves text was never sufficient.

Suite 1865 passed, ruff clean.

cyzus:
@codex review

PR #178, opened 2026-09-02
feat(reminders): time out, parallelise, dedupe and cap reminder providers

cyzus:
P1-2 from the prompt/reminder audit.

## What was wrong

Four gaps in how providers are combined:

| Gap | Effect |
|---|---|
| Global hooks had **no timeout** (only per-turn hooks did) | one hung provider stalled every message indefinitely |
| Providers ran **serially** | each slow one delayed all those after it |
| **No dedupe** | two providers emitting the same text paid for it twice |
| **No ceiling** on the assembled body | unbounded growth into the prompt |

## What I did not do, and why

The audit proposed a structured `ReminderFragment` with `priority`, `max_chars`, `dedupe_key` and `trust` fields. I did not build that.

Declaring per-provider sizes is the same mistake that cost the most review rounds elsewhere in this work: the payload walker enumerated known types and leaked three times; the skill revision fingerprinted selected inputs and missed a path change. **A declaration drifts from what is actually emitted; the assembled string cannot.** So the budget is measured on the final body, and dedupe compares actual text.

Priority is already expressed by registration order, and `gather` preserves it — so that is now the documented, tested contract rather than a new field to keep in sync.

## Truncation choices

**Whole fragments, dropped from the end.** Half a plan snapshot is worse than none, because the model cannot tell it is reading a fragment.

**One oversized fragment is still delivered**, even alone over budget. Truncating to nothing is indistinguishable from a provider that produced nothing, and the caller has no way to notice the difference.

Budget is 6000 characters, the figure the audit suggested. It should be re-derived from real prompt traces — #167 already emits per-section sizes, so the data is there.

## Tests

7 cases: a hung global hook cannot stall the turn, a failing provider does not lose the others, providers run concurrently (0.3s serial vs ~0.1s), identical fragments appear once, the body stays within budget, truncation keeps the earlier fragment intact and drops the later one whole, and a single oversized fragment still arrives.

Six of the seven fail against the previous code — verified by stashing `src/`.

Suite 1850 passed, ruff clean.

### DIFF
--- a/src/suzent/core/system_reminder.py
+++ b/src/suzent/core/system_reminder.py
@@ -843,6 +843,62 @@ def clear_per_turn_hooks() -> None:
 # ---------------------------------------------------------------------------
 
 
+#: How long any one reminder provider may take before it is dropped for the turn.
+HOOK_TIMEOUT_SECONDS = 2.0
+
+#: Ceiling on the assembled reminder body, in characters.
+#:
+#: Measured on the text that actually gets sent, not on per-provider
+#: declarations. Providers that describe their own size drift from what they
+#: emit; the assembled string cannot.
+REMINDER_BUDGET_CHARS = 6000
+
+
+def _dedupe_fragments(parts: list[str]) -> list[str]:
+    """Drop fragments identical to one already present, keeping the first.
+
+    Two providers can produce the same text — a task list surfaced by both the
+    plan hook and an ad-hoc caller, say — and paying for it twice buys nothing.
+    """
+    seen: set[str] = set()
+    unique = []
+    for part in parts:
+        if part in seen:
+            continue
+        seen.add(part)
+        unique.append(part)
+    return unique
+
+
+def _apply_budget(parts: list[str], chat_id: str) -> list[str]:
+    """Keep the assembled body under REMINDER_BUDGET_CHARS.
+
+    Whole fragments are dropped from the end rather than characters from the
+    middle: half a plan snapshot is worse than no plan snapshot, because the
+    model cannot tell it is reading a fragment. Registration order is priority
+    order, so the last-registered providers yield first.
+
+    The first fragment is always kept, even alone over budget. A reminder that
+    truncates to nothing is indistinguishable from a provider that produced
+    nothing, and the caller has no way to notice.
+    """
+    separator_cost = len(FRAGMENT_SEPARATOR)
+    kept: list[str] = []
+    used = 0
+    for index, part in enumerate(parts):
+        cost = len(part) + (separator_cost if kept else 0)
+        if kept and used + cost > REMINDER_BUDGET_CHARS:
+            dropped = len(parts) - index
+            logger.warning(
+                f"[system-reminder] chat={chat_id} over budget: dropped {dropped} "
+                f"of {len(parts)} fragment(s) at {used}/{REMINDER_BUDGET_CHARS} chars"
+            )
+            break
+        kept.append(part)
+        used += cost
+    return kept
+
+
 async def build_combined_reminder(
     chat_id: str,
     deps: Any,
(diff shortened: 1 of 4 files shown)


cyzus:
@codex review

cyzus:
All three confirmed and fixed in `33b30ad2`.

**Measuring raw text (P2, and the one I should be most embarrassed by):** correct. I wrote in this PR's description that the budget must measure *what actually gets sent* rather than per-provider declarations — and then measured the pre-sanitized string. Sanitizing turns each one-character PUA delimiter into `[reminder-delimiter]`, so with user-influenced goal and memory text a body under 6,000 characters can arrive at tens of thousands. Fragments are sanitized before measurement now. The rule was right; I applied it to the wrong string.

**Display trigger unbudgeted (P2):** correct. It is prepended inside the block, so scheduler, heartbeat, goal-continuation and sub-agent turns stayed unbounded despite the cap. It reserves against the same budget now.

**Ad-hoc dropped first (P2):** correct, and the priority was backwards. Caller-supplied reminders are the directives *specific to this turn* — remote-caller attribution, the `analyze_image` instruction for a non-vision model — while hook output is ambient and reproduces next turn. Appending them last meant ambient skill and plan content could crowd out the one instruction the turn depended on. They come first now.

That last one also corrects something I claimed: I said registration order was already an implicit priority and I was making it explicit. It was implicit *and wrong* for ad-hoc reminders, which never went through registration at all.

4 tests, all failing against the previous commit: sanitized measurement, trigger reserved, directives surviving, and the reordering.

Suite 1853 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `f0aacbb9`.

**Dropping the plan fragment (P1):** correct, and it exposes that I treated all fragments as equally droppable. They are not. With skills and repository registered ahead of it, a large catalog could push out the plan fragment — leaving the agent working on a goal it can no longer see. The goal objective is now registered first, so it survives truncation.

On the state-mutation half: `plan_reminder_hook` incrementing `turns_elapsed` on read is removed entirely by #177, which is in review now. This change makes the ordering correct independent of that, so it does not depend on merge order — but the combination is what fully closes your point.

**Blocking providers (P2):** correct, and it means the timeout I added was partly theatre. `asyncio.wait_for` cannot cancel a provider that blocks the loop, because cancelling requires the loop to run — so `plan_reminder_hook`'s synchronous database access could hold the turn open past `HOOK_TIMEOUT_SECONDS` exactly as before. Its body runs on a worker thread now, where the deadline can actually fire, with a test that proves a slow-but-threaded provider does not delay the turn.

I have not converted the repository hook: its work is reading already-loaded `deps.repository_agent_files`, not filesystem access at hook time. If you read that differently I will move it too.

3 tests. Suite 1856 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed, and resolved by **stacking this PR on #177** rather than patching around them — base changed to `fix/goal-turn-count-lifecycle`.

Both findings exist only because `plan_reminder_hook` still mutates `turns_elapsed` while building the reminder:

- a timed-out hook's worker thread keeps running and commits the increment anyway;
- `db.update_goal` from a thread reaches `emit_goal_task_changed`, which is event-loop-affine.

#177 removes that mutation entirely — the hook becomes a pure read — so both disappear rather than needing a guard. Rebasing was the honest fix: adding cancellation checks or loop marshalling around a mutation that is being deleted would have been dead code the day #177 landed.

**Being read-only is also what makes the `to_thread` change safe**, which I had not articulated before. A worker thread that outlives its cancelled await changes nothing if it only reads. The docstring now says that explicitly, so the two properties are recorded as connected rather than as two unrelated decisions.

Please merge #177 first. Suite 1876 passed, ruff clean.

cyzus:
Closing note: GitHub auto-closed this when `fix/goal-turn-count-lifecycle` was deleted on merging #177, and a closed PR cannot be retargeted or reopened. Continued as **#179** — same branch, same commits, rebased onto `main`.

My mistake in sequencing: I should have retargeted this to `main` *before* merging #177, or merged without `--delete-branch`. The review history here is worth keeping, hence the pointer rather than a silent re-open attempt.

PR #179, opened 2026-09-02
feat(reminders): time out, parallelise, dedupe and cap reminder providers

cyzus:
P1-2 from the prompt/reminder audit. **Replaces #178**, which GitHub auto-closed when its base branch (`fix/goal-turn-count-lifecycle`, now merged as #177) was deleted. Same branch, same commits, rebased onto `main`; the review history and my replies are on #178.

## What was wrong

Four gaps in how reminder providers are combined:

| Gap | Effect |
|---|---|
| Global hooks had **no timeout** (only per-turn hooks did) | one hung provider stalled every message indefinitely |
| Providers ran **serially** | each slow one delayed all those after it |
| **No dedupe** | two providers emitting the same text paid twice |
| **No ceiling** on the assembled body | unbounded growth into the prompt |

## What I deliberately did not build

The audit proposed a structured `ReminderFragment` carrying `priority`, `max_chars` and `dedupe_key`. Declared sizes drift from what is actually emitted — the same mistake that cost several rounds elsewhere in this work — so the budget measures the **assembled, sanitized text** and dedupe compares actual strings. Priority is registration order, which `gather` preserves.

## Fixed during review on #178

- **Budget measured raw text** while `wrap_in_system_reminder` sanitizes afterwards, and sanitizing expands each PUA delimiter into `[reminder-delimiter]` — a body under the cap could arrive many times over it. It measures the sanitized text now, which is what the PR claimed from the start and did not do.
- **The display trigger was unbudgeted**, leaving cron, heartbeat and sub-agent turns unbounded.
- **Ad-hoc reminders were dropped first** despite being the turn-specific directives (peer attribution, the `analyze_image` instruction) while hook output is ambient. They come first now.
- **The goal fragment could be crowded out** by a large skill catalog, leaving the agent working on a goal it could not see. Plan is registered first.
- **A blocking provider could not be timed out** at all, because `wait_for` needs the loop to run. `plan_reminder_hook`'s synchronous DB access runs on a worker thread — safe precisely because #177 made that hook a pure read.

## Tests

14 cases covering timeouts, isolation, concurrency, dedupe, the budget, truncation boundaries, sanitized measurement, trigger accounting, directive priority, and blocking providers. Each was checked against the unfixed code.

Suite 1878 passed, ruff clean.

### DIFF
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -14,7 +14,7 @@ import shutil
 import time
 import uuid
 from pathlib import Path
-from typing import AsyncGenerator, List, Dict, Any, Optional
+from typing import AsyncGenerator, List, Dict, Any, Optional, Sequence
 
 from ag_ui.core import (
     CustomEvent,
@@ -798,7 +798,6 @@ class ChatProcessor:
 
         # --- System Reminder Injection (includes per-turn RAG hook when memory enabled) ---
         from suzent.core.system_reminder import (
-            TRIGGER_SEPARATOR,
             build_combined_reminder,
             make_user_prompt_part,
             sanitize_untrusted_text,
@@ -855,9 +854,10 @@ class ChatProcessor:
         stream_run_id = str(uuid.uuid4())
         display_trigger = None
         if system_reminders and not (message_content and message_content.strip()):
-            display_trigger = TRIGGER_SEPARATOR.join(
-                r.strip() for r in system_reminders if r and r.strip()
-            )
+            # The constituents, not a join. build_combined_reminder has to match
+            # these against the same strings arriving as fragments, and a join
+            # cannot be taken back apart once a reminder contains the separator.
+            display_trigger = [r.strip() for r in system_reminders if r and r.strip()]
         # Stateless chats (dream, sub-agents) run a fixed, self-contained prompt and
         # must not receive skill-discovery / plan / RAG reminders — that ambient
         # chatter (e.g. the suzent-automation skill hint) is what made the dream agent
@@ -2722,7 +2722,7 @@ def _trigger_placeholder_prefix() -> str:
 
 
 def trigger_rows_for_snapshot(
-    display_trigger: str | None, is_heartbeat: bool, part: Any
+    display_trigger: str | Sequence[str] | None, is_heartbeat: bool, part: Any
 ) -> list[dict]:
     """Display rows that must become durable with the history they describe.
 
@@ -2741,7 +2741,16 @@ def trigger_rows_for_snapshot(
     if not display_trigger or is_heartbeat:
         return []
 
-    from suzent.core.system_reminder import sanitize_untrusted_text
+    from suzent.core.system_reminder import TRIGGER_SEPARATOR, sanitize_untrusted_text
+
+    # Joined the one way the reminder joins it, so the visible row says what the
+    # model was actually shown.
+    if not isinstance(display_trigger, str):
+        display_trigger = TRIGGER_SEPARATOR.join(
+            r.strip() for r in display_trigger if r and r.strip()
+        )
+        if not display_trigger:
+            return []
 
     # Same treatment wrap_in_system_reminder gives it. Scheduled prompts and
     # subagent results are user-influenced, and _preserve_display_triggers
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -2741,26 +2741,26 @@ def trigger_rows_for_snapshot(
     if not display_trigger or is_heartbeat:
         return []
 
-    from suzent.core.system_reminder import TRIGGER_SEPARATOR, sanitize_untrusted_text
+    from suzent.core.system_reminder import canonical_display_trigger
 
-    # Joined the one way the reminder joins it, so the visible row says what the
-    # model was actually shown.
-    if not isinstance(display_trigger, str):
-        display_trigger = TRIGGER_SEPARATOR.join(
-            r.strip() for r in display_trigger if r and r.strip()
-        )
-        if not display_trigger:
-            return []
+    # The same function the block uses, so the row says what the model was
+    # actually shown. Rendering it independently here is how the two came to
+    # disagree: the block deduplicated repeated reminders and the transcript
+    # kept showing them twice.
+    display_trigger = canonical_display_trigger(display_trigger)
+    if not display_trigger:
+        return []
 
-    # Same treatment wrap_in_system_reminder gives it. Scheduled prompts and
-    # subagent results are user-influenced, and _preserve_display_triggers
-    # prefers this stored content over the placeholder rebuilt from history — so
-    # storing the raw value would put delimiters back into the visible transcript
-    # permanently, undoing the sanitizing on the path that does it.
     stamp = getattr(part, "timestamp", None)
     row: dict = {
         "role": "system_triggered",
-        "content": sanitize_untrusted_text(display_trigger),
+        # Already sanitized by canonical_display_trigger, which is the same
+        # treatment wrap_in_system_reminder gives it. Scheduled prompts and
+        # subagent results are user-influenced, and _preserve_display_triggers
+        # prefers this stored content over the placeholder rebuilt from history
+        # — so an unsanitized value would put delimiters back into the visible
+        # transcript permanently.
+        "content": display_trigger,
         "trigger_origin": "runtime",
     }
     if stamp:
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -14,7 +14,7 @@ import shutil
 import time
 import uuid
 from pathlib import Path
-from typing import AsyncGenerator, List, Dict, Any, Optional, Sequence
+from typing import AsyncGenerator, List, Dict, Any, Optional
 
 from ag_ui.core import (
     CustomEvent,
@@ -900,7 +900,7 @@ class ChatProcessor:
             message_history.append(new_request)
 
             pending_trigger_rows.extend(
-                trigger_rows_for_snapshot(display_trigger, is_heartbeat, parts[0])
+                trigger_rows_for_snapshot(full_prompt, is_heartbeat, parts[0])
             )
 
             # Pre-save the user message to the DB display log for social chats so the
@@ -2722,9 +2722,13 @@ def _trigger_placeholder_prefix() -> str:
 
 
 def trigger_rows_for_snapshot(
-    display_trigger: str | Sequence[str] | None, is_heartbeat: bool, part: Any
+    rendered_prompt: str | None, is_heartbeat: bool, part: Any
 ) -> list[dict]:
     """Display rows that must become durable with the history they describe.
+
+    Takes the rendered prompt, not the reminders that went into it: the row has
+    to say what the model was shown, and the only thing that knows that is the
+    text that was sent.
 
     A reminder-only turn has no user text, so its row cannot be rebuilt once the
     reminder block stops being authenticatable after a restart. The row is keyed
@@ -2738,16 +2742,20 @@ def trigger_rows_for_snapshot(
     someone's visible transcript. An earlier attempt at this shipped exactly that
     bug, so the rule is asserted by a test rather than trusted to a comment.
     """
-    if not display_trigger or is_heartbeat:
+    if not rendered_prompt or is_heartbeat:
         return []
 
-    from suzent.core.system_reminder import canonical_display_trigger
+    from suzent.core.system_reminder import extract_system_reminder_display_trigger
 
-    # The same function the block uses, so the row says what the model was
-    # actually shown. Rendering it independently here is how the two came to
-    # disagree: the block deduplicated repeated reminders and the transcript
-    # kept showing them twice.
-    display_trigger = canonical_display_trigger(display_trigger)
+    # Read back out of the block that was just built, rather than derived from
+    # the same inputs a second time. Every rule the builder applies — joining,
+    # sanitizing, deduplicating, budgeting — is already baked into that text,
+    # and each time this path recomputed one of them it reproduced some and
+    # missed others: the transcript showed reminders twice after the model
+    # stopped seeing them twice, then recorded reminders dropped by the budget
+    # that the model never received at all. There is nothing here left to keep
+    # in step.
+    display_trigger = extract_system_reminder_display_trigger(rendered_prompt)
     if not display_trigger:
         return []
 
--- a/src/suzent/core/chat_processor.py
+++ b/src/suzent/core/chat_processor.py
@@ -798,6 +798,7 @@ class ChatProcessor:
 
         # --- System Reminder Injection (includes per-turn RAG hook when memory enabled) ---
         from suzent.core.system_reminder import (
+            TRIGGER_SEPARATOR,
             build_combined_reminder,
             make_user_prompt_part,
             sanitize_untrusted_text,
@@ -854,7 +855,7 @@ class ChatProcessor:
         stream_run_id = str(uuid.uuid4())
         display_trigger = None
         if system_reminders and not (message_content and message_content.strip()):
-            display_trigger = "\n\n---\n\n".join(
+            display_trigger = TRIGGER_SEPARATOR.join(
                 r.strip() for r in system_reminders if r and r.strip()
             )
         # Stateless chats (dream, sub-agents) run a fixed, self-contained prompt and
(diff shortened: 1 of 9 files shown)


cyzus:
@codex review

cyzus:
Both confirmed and fixed in `49381133`.

**Blocking before the first await (P1):** correct, and I had checked the wrong hook. When you raised this on #178 I looked at `repository_agents_reminder_hook`, saw it reading `deps.repository_agent_files`, and concluded it did no filesystem work — I missed the `Path.resolve()` inside the loop, and said so on the PR. It resolves every discovered agent file, so an unreachable network mount stalls the turn regardless of the timeout. It runs off the loop now.

**Unbounded threads (P2):** correct, and it means my previous fix was weaker than I described. I moved the plan hook to `asyncio.to_thread` and called the deadline enforceable; `to_thread` uses the shared default executor, so a wedged SQLite occupies workers that every other `to_thread` caller in the process depends on. Read-only prevents late writes, which is what I argued, but it does nothing about resource exhaustion.

Both providers now use a dedicated four-thread pool. Being plain about the limit rather than overstating it again: **this does not make the work cancellable.** Python cannot kill a thread. What it does is keep the turn from waiting, and confine stragglers to a pool nothing else uses. Cancellable async I/O for SQLite and path resolution would be the real fix and is well beyond this PR.

4 tests: both providers use the pool and neither uses `to_thread`, the pool is separate and bounded, and a wedged provider does not delay the turn.

Suite 1881 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `ce3e4365`.

**Trigger charged twice (P2):** correct. A reminder-only turn passes the same text as `display_trigger` *and* as an ad-hoc fragment, and the trigger is prepended inside the block — so the model read it twice and paid for it twice, and my first-fragment exemption waved the duplicate straight through the cap. I fixed the symptom you offered as the alternative (checking the first fragment against the remaining budget) in an earlier round; this time I removed the duplication itself, which is the actual cause.

That change then broke something on its own: dropping the only fragment empties the body, and the existing early return would have discarded the entire reminder — losing the visible record of a turn whose *only* record it is. Caught by the test I wrote for the first fix returning `None`. Guarded now, with its own test.

**Executor blocks shutdown (P2):** correct, and a good catch on a consequence I had not considered at all. `ThreadPoolExecutor` joins its workers at interpreter exit, so exactly the wedged read this pool exists to survive would hang every server stop, reload and resource-guard recycle — the turn returns in two seconds and the process never exits. Replaced with bounded daemon threads.

At saturation a provider is now skipped rather than queued, deliberately: queueing behind wedged work is how a transient stall becomes permanent. Still not cancellable — Python cannot kill a thread — but the process can now leave, which was the actual gap.

6 tests: daemon threads and no executor, saturation skips, slots released on success and on failure, trigger charged once, and a trigger-only turn still producing a reminder.

Suite 1886 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `562abca2`.

Correct, and the framing is the useful part: this is the third filesystem-reading provider, and the previous two rounds each moved exactly one. I have now made the same mistake three times in the same file — enumerate the paths I can see, ship, discover the one I didn't.

So the guard is over the set rather than over the hooks I remembered:

```python
for hook in (repository_agents_reminder_hook, skills_reminder_hook, plan_reminder_hook):
    assert "run_provider_blocking" in inspect.getsource(hook)
```

Plus a behavioural test that a skill whose `path.resolve()` blocks for 1.5s leaves the loop responsive (ticker keeps running) — that one fails on the unfixed code, verified by reverting the fix and re-running.

**On why the framework doesn't just do this for every hook:** I considered making `_run` offload unconditionally, which would close the class properly rather than by convention. It cannot, cheaply — hooks are coroutines, so offloading means driving them in a thread's own event loop, and anything loop-affine inside one (db handles, deps state, third-party hooks registered outside this repo) would break in ways that are much worse than a stall. So the contract stays "a hook that touches the filesystem routes it through `run_provider_blocking`", and the test above is what makes the contract checkable instead of remembered.

Suite 1888 passed, ruff clean.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `c3bd8ff7`.

You're right, and the arithmetic is the part I should have done myself: three registered blocking hooks against four process-wide slots means **two concurrent turns already exceed the pool**. That isn't an edge case, it's the normal state of a server with more than one active chat — so the guard I added to protect against a wedged provider was itself silently dropping repository, skill and plan context from perfectly healthy requests.

My stated reason for skipping ("queueing behind wedged work turns a transient stall into a permanent one") was sound about *wedged* work and wrong as a blanket policy, because it treated "busy" and "wedged" as the same state. They are distinguishable by the thing already in the code: time.

So the acquire now waits:

```python
while not _provider_slots.acquire(blocking=False):
    await asyncio.sleep(_SLOT_POLL_SECONDS)
```

It yields rather than blocking — a blocking acquire would block the loop, which is the exact failure `run_provider_blocking` exists to prevent — and it is bounded by the caller's existing deadline, so a genuinely wedged pool still fails the provider at `HOOK_TIMEOUT_SECONDS` instead of queueing behind the block forever. That keeps the original concern intact without paying for it on every busy turn.

Three tests, all verified failing against the previous commit: a busy pool waits and then succeeds once a slot frees; the wait ends on the caller's deadline when it never does; and the loop stays responsive throughout (ticker keeps running).

Suite 1890 passed, ruff clean.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `bc535629`.

**Joined trigger (P2):** correct, and it is the *same* coupling that produced the original duplication rather than a new one. `chat_processor` owns the join and `system_reminder` re-guessed it, so my exact-equality filter matched the whole and missed every constituent — with two reminders the model read both twice, and the first-fragment exemption walked them past the cap exactly as you describe.

Fixing the one-element case and calling it done was the error. The separator now has a single definition that both the join and the split import, because two copies of that string is precisely what let them disagree. There is a test asserting the literal does not reappear at the join site.

**Slot leak on thread-start failure (P2):** correct, and worse than it first looks — it is monotonic. The release lives in the worker's `finally`, so a thread that never starts never gives its slot back; four of those and the pool is gone for the life of the process, at which point every provider waits out its full deadline and plan, skill and repository context vanish from every subsequent turn. A transient resource shortage becomes permanent context loss. Wrapped, with a test that patches `Thread` to raise and asserts the count is unchanged.

Three tests, all verified failing against the previous commit.

Suite 1893 passed, ruff clean.

Worth saying plainly: this is round five, and rounds three, four and five each found a real defect in the previous round's fix. Every one has been a case of fixing the instance in front of me instead of the class — the third filesystem hook, busy-versus-wedged, and now single-versus-joined trigger. The guards I have added since are properties rather than instances, which is the right correction, but the pattern itself is worth flagging rather than quietly iterating past.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `87e31e5f`, and I found a second under-count next to it.

**The exemption:** correct. `if kept and ...` never checks the first fragment at all, so with a trigger reserved the cap is exceeded in exactly the situation it was added for. The exemption exists so an empty body cannot be mistaken for a provider that produced nothing — and when there is a trigger, something is already being delivered, so it is now spent by the trigger and every fragment behind it is held to the cap. Without a trigger the exemption behaves as before.

**Found while testing it:** `trigger_cost` measured the trigger's bare text, but it is sent wrapped in `<system-reminder-display-trigger>` tags — about 70 uncounted characters, enough that a 3,000-character trigger and a 3,000-character fragment cleared the cap even after your fix. The envelope now has one definition that both the wrap and the budget use, the same shape as the separator fix last round.

**On why this kept recurring:** every previous version of this check measured the fragment list, which agreed with itself while the text actually sent was over. The new test measures the wrapped output — trigger, envelope and all — across six size combinations including your 5,900+500 case. That is the post-condition; the fragment arithmetic is just how it is met.

Suite 1900 passed, ruff clean. Verified failing against the previous commit (3 of 6 cases).

This is round six, and rounds three through six each found a real defect in the prior round's fix. I have said I would step back rather than patch a seventh time, so: if this one does not come back clean, I will stop and reconsider the budget design as a whole instead of continuing.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `82d204b3` — and this one I fixed as a design change rather than another patch, which is what I said I would do if round seven found something.

You are right, and the general form matters more than the case: **splitting is not the inverse of joining** for any input that contains the separator, so no amount of care at the split site can recover the boundaries. Last round I shared the separator constant between the two sites, which narrowed the gap and left the actual defect in place — the two sites agreed on the string and still could not agree on the boundaries.

So the reconstruction is gone. `build_combined_reminder` now takes the constituents and does the joining itself, because the module that has to deduplicate is the one that has to agree with the join:

```python
display_trigger: Union[str, Sequence[str], None]
```

A single string is one constituent, not a join to be taken apart. `chat_processor` passes its list and no longer joins; `trigger_rows_for_snapshot` joins the one way the reminder does, so the visible row still says what the model was shown.

Three tests: a reminder containing the separator charged once, a single string treated as one constituent, and a source-level check that the caller no longer joins. Two of the three verified failing against the previous commit.

Suite 1902 passed, ruff clean.

Worth recording, since it took seven rounds: the recurring fault was not the individual bugs but reconstructing structured input by parsing rendered output — the fragment-list-versus-sent-text mismatch in round six and the join-versus-split mismatch here are the same mistake in two places. The current shape has no parse step to get wrong.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `f20343a5`, along with a CI failure that landed on the same commit.

**Repeated constituents (P2):** correct. The fragments are deduplicated and the trigger was not, so one reminder passed twice went out in full twice — and at 4,000 characters each that clears the cap on its own through the oversized-trigger exemption. Now deduplicated in order through the same `_dedupe_fragments` the fragments use, which is the point: it is the same rule, so it should not be a second implementation of it.

**Backend CI failed on `82d204b`** — `assert 3 == 4` on the slot count, in two tests. Not a flaky test: the worker released its slot in `finally` *after* `call_soon_threadsafe`, so the awaiter could resume while the thread still held it. The pool read as short by one for the length of the handoff — briefly under-subscribed under load, and a genuine race for anything checking capacity the moment a provider returns. Releasing before the wake makes the ordering deterministic; the work is finished by then, so there is nothing to hold it for.

Two tests, and the slot tests pass on repeated runs rather than once.

Suite 1904 passed, ruff clean, and I will not merge until Backend is green on this commit.

cyzus:
@codex review

cyzus:
Both confirmed and fixed in `5dbf0b0e`, and they turned out to be one cause on each side of the same line.

**Sanitizer-induced collision:** correct. Deduplicating compared the *inputs* when the thing that decides sameness is the *delivered* text — a delimiter and its literal `[reminder-delimiter]` spelling differ going in and are identical coming out, so both survived and the reminder went out twice. Sanitizing now happens before deduplicating, which is the only order in which the comparison means anything.

**Transcript row:** correct, and it is the second rendering I left behind last round. I moved the join into the module precisely so the two sides could not disagree, then had `trigger_rows_for_snapshot` keep rendering its own — so the model saw one copy and the stored row kept showing two.

Both now go through `canonical_trigger_constituents` / `canonical_display_trigger`. The row no longer knows what the separator is, and a test asserts it does not reacquire the knowledge.

Three tests, all verified failing against the previous commit.

Suite 1907 passed, ruff clean, Backend green on the previous commit.

Round nine, and the last two rounds have both been *this*: a rule applied to the model-visible path and not to the transcript path, or applied to inputs and not to delivered text. The single-definition move each time is right and I keep making it one call site too late.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `35911227`.

Correct. Joining before budgeting made the whole trigger a single item, and a single item is precisely what the oversized exemption waves through — so N unrelated 4,000-character reminders cleared the cap together. Budgeting the constituents before the join means the exemption covers one reminder, which is the most that genuinely cannot be helped.

**A second thing fell out of implementing it, and it is the more useful one.** My first attempt reserved the trigger's envelope in the constituents pass, which — because the exemption was *inferred* from `reserved > 0` — silently withdrew the exemption and dropped a lone oversized reminder entirely, losing the whole turn. Reserving 71 characters for tags is not the same event as having delivered content, and the inference conflated them. It is now an explicit `exempt_first` parameter.

I want to be straight about how that was caught: not by reasoning about the change, but by the test I had written for the previous round's behaviour failing. Every one of the last several rounds has had that shape.

Five tests here (three sizes, plus the single-oversized case both with and without a trigger), verified failing against the previous commit.

Suite 1911 passed, ruff clean.

This is round ten. Rounds three through ten have each found a real defect, and four of them — this one included — were in the budget's exemption logic specifically. The individual fixes have converged (the last three were single-definition moves that each held), but I am no longer confident that iterating is the right process for this particular area rather than a sign the exemption concept itself is carrying too much. I have said as much to the repository owner and am pausing the automatic review loop here rather than continuing to round eleven unprompted.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `7f87bcf3` — and this time by removing the derivation rather than adding the missing rule to it.

You are right: canonicalization handled duplicates and did not reproduce `_apply_budget`, so the transcript recorded reminders the model never received. But adding budgeting to that path would have been the third time I taught it one of the builder's rules, and there would have been a fourth.

The row now reads the trigger back out of the **rendered prompt** — the text that was actually sent — via `extract_system_reminder_display_trigger`. Joining, sanitizing, deduplicating and budgeting are all already baked into it. There is no second derivation left to keep in step, which is the only version of this fix that does not have a next round.

The existing snapshot tests now build a real reminder and pass the rendered prompt, so they exercise the sent text instead of a stand-in. New test: two 4,000-character reminders, the first recorded and the second absent because the budget dropped it — verified failing against the previous commit.

Suite 1912 passed, ruff clean.

Round eleven. For the record on this PR: rounds 3–11 each found a real defect, and they fall into two families — the budget exemption (4) and the model-visible path versus the transcript path (3). Both families are now closed structurally rather than case by case: the exemption is an explicit parameter applied to one item, and the transcript path has no independent derivation at all.

cyzus:
@codex review

cyzus:
Confirmed, and fixed by removing the exemption rather than correcting it — the repo owner picked truncation when I put the trade-off to them.

You are right about the window: between ~5,930 and 6,000 characters an item passes the bare check and renders over the cap once its ~70-character tag envelope is added.

I did not take the suggested fix, though, and the reason is worth stating. Exempting only items larger than the *full* budget means a 6,000-character reminder is dropped, and on a reminder-only turn that reminder is the entire turn — so the cap would be honoured by delivering nothing at all, silently. That is a worse failure than 69 characters of overshoot.

There is no exemption now. An item too large is cut to what is left and marked:

```
[truncated: over the system-reminder budget]
```

Sending it whole is the cap failing; dropping it loses the delivery silently; cutting it loses the tail and says so, which is the only one of the three the model can react to. Whole fragments are still dropped from the end rather than trimmed — half a plan snapshot is worse than none — so only the first item, and only when nothing else would survive, is ever cut.

**On the five rounds this makes.** Every one of them was an item claiming to be the single thing that could not be helped: the joined trigger, the first fragment behind a delivered trigger, duplicate copies, distinct constituents, and now the envelope. Each fix narrowed the exemption and the next round found another way in. The concept was the defect — an escape hatch has to be defined by what it excludes, and nothing ever defined it. There is now no code path that skips the cap check, asserted by a test.

Nine tests, including the exact 5,940/5,999/6,000/6,001/7,000 boundary and an assertion that the turn's only content never vanishes. Verified failing against the previous commit.

Suite 1918 passed, ruff clean.

Follow-up filed as cyzus/suzent#180: the truncated tail should be spilled to a temp file the agent can read, and the same policy should cover tool output. Deliberately out of scope here.

cyzus:
@codex review

cyzus:
Confirmed and fixed in `bbaad2ec`.

Correct, including the mechanism: the first reminder leaves less room than the separator costs, the second is dropped from the trigger, and it then fits as a fragment precisely because a fragment pays no separator when it is first.

The filter was built from the constituents that *survived* budgeting, when the question it answers is which strings the caller also handed in as fragments — and the trigger's budget has no bearing on that. It is now built before budgeting. Truncation had the same hole, since a cut constituent no longer equals the fragment it came from; the same change closes both.

Reproducing this took me two attempts. My first test used round numbers, passed against the unfixed code, and proved nothing — the window is about seven characters wide, so I computed the boundary from `render_trigger_block` and `TRIGGER_SEPARATOR` rather than guessing at sizes. The test derives them rather than hard-coding, so it stays on the boundary if either changes.

Verified failing against the previous commit, and the assertion is that the block and the row agree on the outcome rather than that any particular outcome occurs.

Suite 1919 passed, ruff clean.

cyzus:
@codex review

PR #235, opened 2026-09-18
fix(web): stop dropping a finished turn when its reply is short

cyzus:
A stream could finish, render nothing, and only appear after a manual refresh. The turn was in the database the whole time — the frontend threw the snapshot away.

## What was happening

Three guards in `loadChat` compare "the last assistant message" on each side to decide whether a server snapshot is stale. They were written when the frontend always appended optimistically and raced the backend.

Since the snapshot/cursor protocol (#227), the confirmed path deliberately appends nothing locally — the backend is authoritative. So the local store's last assistant belongs to the **previous** turn, and the guards were comparing two unrelated messages. A new reply shorter than the previous one, or one still ending in a tool call, looked like a regression, and the entire snapshot was rejected. `clearParts()` then wiped the transient stream, leaving a blank chat until the user refreshed.

## What changed

**Anchor the guards to one turn.** `pairLastTurnAssistants` pairs the local store's last assistant with the server's assistant for that same turn (turn index = how many user messages precede it). A turn the local store has never seen is new information, not a regression. The "server hasn't written this turn yet" case still keeps local content.

Applied to all three:
- `shouldKeepLocalAssistantContent`
- the pending-approval guard — this one was worse than its comment claimed: it scanned the *entire* server history with `.some()`, so pending state left by an old session could reject any snapshot
- the shorter-content guard (`serverContent.length + 40 < localContent.length`)

**Skip the guesswork where the backend has already answered.** `/chat/live` only emits `STREAM_END` after post-processing reports success (`stream_replay.py`), and `recoverableStream` throws on anything else — so a stream reporting `confirmed` has a complete turn in the DB. `loadChat` gains `trusted` for that case, sharing the existing `authoritative` bypass. Both confirmed callers use it: the direct send path and the live/background cleanup path (cron, social, heartbeat, reconnect-after-navigation).

The guards still run on paths with no persistence signal, where the optimistic content they protect is real.

## Tests

Three cases added to `chatSyncGuards.test.ts`: the short-reply regression, a new turn that is still tool-only, and the case that must keep behaving as before (server has not written this turn at all). Full frontend suite: 386 passed.

## Follow-ups (not in this PR)

- `useAGUI` still carries `steerStream`, `resumeStream` and the `formData` multipart branch. All three are unreachable — steer and approval-resume moved to `/chat/steer-send` and `/chat/send`, and no caller passes `formData`. ~150 lines of a previous transport generation.
- Stop/abort is the remaining path with no persistence signal: `stopStreaming` clears parts and bets on `setTimeout(loadChat, 500)`, with no retry. Same blank-until-refresh failure, different trigger.
- ACP runs attach no `replay.persistence`, and `persisted` returns `True` when it is `None`. Correct today only because ACP persists synchronously before `AGENT_FINISHED` — an implicit invariant worth making explicit.


cyzus:
Good catch — fixed in 
07b512a03.

`canvas_action` (local-only pill, persisted as `user`) and `system_triggered` (cron/heartbeat) now advance the turn counter alongside `user`; `notice` still doesn't, since it never prompts a response. Both cases have a regression test, and both tests fail against the previous `user`-only counter.

cyzus:
@codex review

